Issue #78, opened 2018-01-10
Daemonset/sidecar container?


raghur:
So at the outset, I have only basic understanding of appinsights's data model. However, I'm currently working on porting a relatively mid sized codebase to .net core with the aim to run it on k8s. Also, I've just landed on your repo from google and a quick read through isn't very informative - however, it did bring up a question. 

The question I have is that would it be possible to explore running this as a sidecar/daemonset in k8s - so that the actual code doesn't have to deal with keeping a buffer/flush it's log etc. Even better if this can also take an [eventflow config](https://github.com/Azure/diagnostics-eventflow) and be responsible for pushing AI telemetry reliably.

xiaomi7732:
@raghur,
Thanks for bringing the question up.

Firstly, let me see if I understand what you are asking here. Are you trying to see if we can host **something** as a service in the sidecar, so that your application throws whatever the events interested to the sidecar? The sidecar handles the communication with application insights server.

If I got that correct, the short answer would be no, we don't support sidecar hosting at this point. However, could you please explain your diagnose requirements, like what are the info that your interested if there's a sidecar running?

Thanks again for your asking.

raghur:
@xiaomi7732 what I'm trying to as is something like this:
 >Firstly, let me see if I understand what you are seeking. Are you trying to see if we can host something as a service in the sidecar, so that your application throws whatever the events interested to the sidecar? The sidecar handles the communication with application insights server.

Yes 

> like what are the info that your interested if there's a sidecar running?

So one thing that I'd love to see is basically node information (cpu/memory etc) correlated with application events. today to achieve that, I'm relying on what K8S downward api provides (node name etc) to add on as additional kv pairs but the info is limited and I don't want to write any code for that. It'd be great if something like this can a) recieve log events from the app, b)additionally enrich/filter/transform  using eventflow diagnostics and c) enrich with node info so that application events can be correlated with node events in AI. 

I recently [asked about APM options](https://www.reddit.com/r/dotnet/comments/7ljuvt/apm_for_dot_net_core_app_on_kubernetes/) for .net core apps on k8s on reddit. That's sort of what I'm after and basically describes my diagnostic requirements.

xiaomi7732:
@raghur, thank you for providing the details.
In short term for this project, we shall be able to provide the CPU/memory info on app events per https://github.com/dotnet/standard/issues/327.

[...]

PR #84, opened 2018-01-17
Add CPU/Memory metrics as custom dimentions

xiaomi7732:
Thrown in a quick way to provide CPU/memory metrics when AI event happens.
A quick dirty solution for #78 .

E2E tested:
```
trce: Microsoft.ApplicationInsights.Kubernetes.KubernetesTelemetryInitializer[0]
      {"Timestamp":"2018-01-17T22:15:01.2764743+00:00","Sequence":null,"Context":
...

"Process.Memory":"104MB","Process.CPU(ms)":"1920"
...
```


karolz-ms:
Have you considered reporting these as AI metrics, as opposed to custom telemetry properties?

brahmnes:
I agree with Karol in that I don't think we should add these metrics on every single telemetry events. Looking at the requirement from https://github.com/Microsoft/ApplicationInsights-Kubernetes/issues/78, it can be probably be achieved by making sure the telemetry events, and the AI metrics, are both decorated with container name as properties, so queries can correlate them.

Having said that, I am not sure reporting CPU/memory metrics from this nuget is the best solution. At a minimum, I would prefer to make this configurable so it's not always collected automatically. An alternative would be to run an agent like telegraf and have it send data to AI.

xiaomi7732:
@karolz-ms,
Thanks for the ask 👍. Yes, AI metrics is the first consideration but I chose to fall back to put these on the telemetry property intentionally.

Firstly, I think **it totally maps** for metrics to be tracked by AI metrics;

Then, there I see the **mismatch** for between AI events in AI.K8s and the profiling style AI metrics. At this point, I don't think it justified to implement a profiling logic in this library based on the ROI.

Third, this still addresses one scenario for those users that want to get basic metrics info and correlate it back to app events in one shot.

Based on those reasons, I think, why don't we just quickly throw in some data on the telemetry? At least, when the event happens, we can get a quick feeling on how is the CPU load and memory load looks like.

Does this address your question? Please let me know what's on your mind. :-)

karolz-ms:
@xiaomi7732 it is a best practice to have a distinction between logs and metrics. They are often going through different pipelines and are handled by different backends. That said I understand this is somewhat of an experimental change, so let's experiment :)

PR #87, opened 2018-02-04
Service collection implementation

xiaomi7732:
A simple refactor to align with .NET Standard dependency injection coding model. #86.

### DIFF
--- a/src/ApplicationInsights.Kubernetes.csproj
+++ b/src/ApplicationInsights.Kubernetes.csproj
@@ -50,18 +50,15 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)'=='netstandard1.3'">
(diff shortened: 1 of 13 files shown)


src/ApplicationInsights.Kubernetes.csproj
```diff
@@ -51,15 +51,18 @@
 
   <ItemGroup Condition="'$(TargetFramework)'=='netstandard1.3'">
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.1.2" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="1.1.2" />
```
karolz-ms: If you depend on Ms.Ex.Logging, do you still need Logging.Abstractions?

xiaomi7732: @karolz-ms good eye 👍 I'll fix it before the merge.

Issue #85, opened 2018-01-19
Investigate using IHostingStartup for ASP.NET 2.0 Core app to inject AI.K8s


xiaomi7732:
http://kalapos.net/Blog/ShowPost/IHostingStartupInAspNetCore20

xiaomi7732:
- [x]  Doable
- [ ] Add implementation
- [ ] Add instruction

PR #88, opened 2018-02-04
Implement IHostingStartup to support ASP.NET Core 2.0 DependencyInjection feature light up

xiaomi7732:
Adding support for Dependency Injection to light up AI.K8s. This will allow the developers to use environment variable than code to light up ApplicationInsights.Kubernetes.
Address #85


xiaomi7732:
@karolz-ms Thanks for the quick turn-aournd :-)

Issue #97, opened 2018-03-01
Keep the application insights k8s working when the app is not running in K8s environment


xiaomi7732:
If we could emit some test info instead of totally stop working, it will help debugging.

Issue #96, opened 2018-02-28
Set proper internal sdk version for telemetry objects


xiaomi7732:
The current SDK version is general that not providing proper info to trace back to this SDK.

xiaomi7732:
Since we are not providing telemetryClient, it seems the most reasonable place to provide the SDK version is through the the TelemetryInitializer. Reference:
https://github.com/Microsoft/ApplicationInsights-Home/issues/87

PR #98, opened 2018-03-05
Feature/sdk version

xiaomi7732:
Address #96  & #97 .
Mainly added the SDK Version for analysis. Use the ability to send stub telemetry data to do the validation.

### DIFF
--- a/src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesModule.cs
+++ b/src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesModule.cs
@@ -4,6 +4,7 @@ using System.Reflection;
 using System.Threading.Tasks;
 using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.ApplicationInsights.Kubernetes.Stubs;
+using Microsoft.ApplicationInsights.Kubernetes.Utilities;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
@@ -80,7 +81,9 @@ namespace Microsoft.ApplicationInsights.Kubernetes
                 if (k8sEnv != null)
                 {
                     // Inject the telemetry initializer.
-                    ITelemetryInitializer initializer = new KubernetesTelemetryInitializer(k8sEnv, serviceProvider.GetService<ILogger<KubernetesTelemetryInitializer>>());
+                    ITelemetryInitializer initializer = new KubernetesTelemetryInitializer(k8sEnv,
+                        SDKVersionUtils.Instance,
+                        serviceProvider.GetService<ILogger<KubernetesTelemetryInitializer>>());
                     configuration.TelemetryInitializers.Add(initializer);
                     logger?.LogDebug("Application Insights Kubernetes injected the service successfully.");
                 }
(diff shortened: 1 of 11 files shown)


src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesModule.cs
```diff
@@ -47,12 +50,15 @@ public static void Initialize(TelemetryConfiguration configuration, TimeSpan? ti
         /// </summary>
         /// <param name="loggerFactory"></param>
         /// <param name="timeout"></param>
-        public static void EnableKubernetes(IServiceCollection serviceCollection, TelemetryConfiguration configuration, TimeSpan? timeout = null)
+        public static void EnableKubernetes(IServiceCollection serviceCollection,
+            TelemetryConfiguration configuration,
+            TimeSpan? timeout = null,
+            bool useStub = false)
```
brahmnes: In general, it's always better to minimize of test specific code in product code.
The most common practice is abstracting with interfaces. Instead of bool useStub, it's better to just accept a IK8sEnvironmentFactory as a parameter defaulting to null. Then create a production default one if it's null. Then the rest of your class can be freed from test specific code like "if (useStub) { "

src/ApplicationInsights.Kubernetes/Utilities/SDKVersionUtils.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Reflection;
+
+namespace Microsoft.ApplicationInsights.Kubernetes.Utilities
+{
+    internal sealed class SDKVersionUtils
+    {
+        #region Signleton
+        private SDKVersionUtils() { }
+        private static readonly SDKVersionUtils _instance = new SDKVersionUtils();
+        public static SDKVersionUtils Instance => _instance;
```
brahmnes: Why not just remove _instance and make Instance a readonly field?

src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesModule.cs
xiaomi7732: @brahmnes , thanks for the review. Great thinking.

Here are some of my perspectives:
Minimize test specific code in production code is very reasonable because we don't want any test code interference the production code. At the best, we don't have test code in production.

However, given the advantage to debug the code without setting up a real K8s environment, that is a compromise I am trying to make here.

Now that we have stub class v.s. the real classes, the risk of misuse shows up. 
And the way to ensure correct service is injected is: to introduce a switch explicitly. That switch could be:
1. IK8sEnvironmentFactory or
2. a boolean
I intentionally had chosen 2 here because: we have 2 stubs needed instead of 1:
* IKubeHttpClientSettingsProvider
* IK8sEnvironmentFactory
And when stubs are used, they are always applied coupled, that is why a single boolean is picked to control the use of the stubs or not. Here we also are not considering allowing 3rd party implementations being injected.

Now that we have picked the boolean as the way, to limit the number of possible point of failures become important. And the place recognized is the injector, where all services got registered. Once the correct service is registered, it will not become wrong. That is where one and only one **if-else** happens.

src/ApplicationInsights.Kubernetes/Utilities/SDKVersionUtils.cs
xiaomi7732: When a member is public, it is a property than a field. I can list the reasons but I don't want to run into the ever long-haul debates here.

xiaomi7732:
Abandon this one.

Issue #94, opened 2018-02-27
Put a hard dependency on ApplicationInsights.ASPNet for self-host startup


xiaomi7732:
Self-host startup only works in ASP.NET Core 2.0. Adding the dependency will making sure the requirement AI component will be available.

PR #101, opened 2018-03-11
Enable Applicaiton Insights for hoststartup scenario

xiaomi7732:
Fix for #94


xiaomi7732:
Thanks for the quick review!

PR #108, opened 2018-04-17
Examples/basic scenario

xiaomi7732:
Providing a running scenario for the example. This walks the user for the very basic configuration to make the library work with their ASP.NET 2.0 application in Kubernetes.

* If this turns out good, I'll update the main README.md to pointing to this example.
* This also helps quickly manual verify bug fixes.


examples/BasicUsage/README.md
```diff
@@ -0,0 +1,52 @@
+# Walkthrough
+This example walks through the necessary steps to create an ASP.NET Core 2.0 MVC application to Kubernets cluster with `Application Insights for Kubernetes` on.
```
karolz-ms: ... DEPLOY an ASP.NET Core 2.0 MVC application to Kubernetes cluster ...?

examples/BasicUsage/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+_Note: This is an example, not a best practice. You will need to consider more around the security aspects like how to protect the application insights instrumentation key and so on._
```
karolz-ms: How about the following:

This is a simple example that does not follow all best practices, including security-related best practices. E.g. Application Insights instrumentation key is not adequately protected (it should be deployed as a secret)

examples/BasicUsage/README.md
```diff
@@ -0,0 +1,52 @@
...
+    services.EnableKubernetes();
+    services.AddMvc();
+}
+```
+* Optionally, update the base images:
+```
+docker pull microsoft/aspnetcore-build:2.0
+docker pull microsoft/aspnetcore:2.0
+```
+* Build the docker container (saars/aik8sbasic, for example) using [Dockerfile](Dockerfile) and upload it to an image registry.
```
karolz-ms: I would make the Docker tag more abstract e.g. `your-image-registry/aik8sbasic:latest` 

As much as we all like you, Saar, I am not sure how many users will have a registry named `saars`  :-)

examples/BasicUsage/k8s.yaml
```diff
@@ -0,0 +1,34 @@
...
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: webapp
+    spec:
+      containers:
+      - name: ai-k8s-basic-container
+        image: saars/aik8sbasic:latest
```
karolz-ms: See my comment about image registry name above

examples/BasicUsage/README.md
xiaomi7732: 🥇  They won't be able to push to my docker hub repository but I'll update it :-)

Issue #109, opened 2018-04-17
Follow the application insights pattern to inject the proper module into the service collection


xiaomi7732:
Details:
https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/develop/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs

PR #111, opened 2018-04-18
Dev/saars/refactor init module

xiaomi7732:
Address #109 .
Basically, the idea is: allow the TelemetryModel and the ITelemetryInitializer finish injected to the service collection. Spin up the query for the K8s properties but delay do NOT be blocked by the result.
When there's no result and not timeout yet, skip the initializer for the telemetry until:
1. there's K8s info;
2. timeout.
When it times out, log an error.

### DIFF
--- a/src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
+++ b/src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
@@ -33,7 +33,7 @@ namespace Microsoft.ApplicationInsights.Kubernetes
         /// Async factory method to build the instance of a K8sEnvironment.
         /// </summary>
         /// <returns></returns>
-        public async Task<K8sEnvironment> CreateAsync(TimeSpan timeout)
+        public async Task<IK8sEnvironment> CreateAsync(TimeSpan timeout)
         {
             K8sEnvironment instance = null;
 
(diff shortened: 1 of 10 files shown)


src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
```diff
@@ -33,7 +33,7 @@ internal class K8sEnvironmentFactory : IK8sEnvironmentFactory
         /// Async factory method to build the instance of a K8sEnvironment.
         /// </summary>
         /// <returns></returns>
-        public async Task<K8sEnvironment> CreateAsync(TimeSpan timeout)
+        public async Task<IK8sEnvironment> CreateAsync(TimeSpan timeout)
```
karolz-ms: For the timeout parameter to really have the effect on the whole operation, all async calls that are part of it should consider the timeout too. If a method does not take a timeout parameter, you can use Task.WaitAny() with a second task being Task.Delay(remainingTime). I understand this is not the main focus of your change though.

tests/UnitTests/KubernetesModuleTests.cs
```diff
@@ -14,7 +14,7 @@ public class KubernetesModuleTests
         public void ServiceInjected()
```
karolz-ms: What about tests that verify:

1. That the timeout is effective?
2. That the initializer will not add any properties and will not crash when K8s data cannot be retrieved?

Issue #110, opened 2018-04-17
Enable signing the NuGet package


xiaomi7732:
Sign the NuGet package to enhance the security of the package.

PR #112, opened 2018-04-19
Reference MicroBuild 0.3.0 to support nuget package sign

xiaomi7732:
Address #110 .
NuGet package sign requires MicroBuild 0.3.0 and above to works for multi-targeting project like the ApplicationInsights.Kubernetes package.


SergeyKanzhelev:
Is version upgrade the only thing needs to happen? I'm asking to move the same approach into other Application Insights repositories.

xiaomi7732:
@SergeyKanzhelev I'll forward you an internal mail.

Issue #105, opened 2018-04-11
Support Windows Containers


david-jarman:
Using this package in a Windows container will cause the application to throw an exception on startup:

```
System.IO.FileNotFoundException: File contains container id does not exist.
File name: '/proc/self/cgroup'
   at Microsoft.ApplicationInsights.Kubernetes.KubeHttpClientSettingsProvider.FetchContainerId(String pathToCGroup)
   at Microsoft.ApplicationInsights.Kubernetes.KubeHttpClientSettingsProvider..ctor(ILogger`1 logger, String pathToToken, String pathToCert, String pathToNamespace, String pathToCGroup, String kubernetesServiceHost, String kubernetesServicePort)
   at Microsoft.ApplicationInsights.Kubernetes.KubernetesModule.<>c.<BuildK8sServiceCollection>b__3_0(IServiceProvider p)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitScoped(ScopedCallSite scopedCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitScoped(ScopedCallSite scopedCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.<>c__DisplayClass22_0.<RealizeService>b__0(ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
   at Microsoft.ApplicationInsights.Kubernetes.KubernetesModule.EnableKubernetes(IServiceCollection serviceCollection, TelemetryConfiguration configuration, Nullable`1 timeout)
```

Since there is no cgroups concept in Windows, the container id must be retrieved a different way. A potential solution for docker containers is to use the "computername" environment variable. It is set to a shortened version of the container id.

xiaomi7732:
@david-jarman,
[...]

PR #113, opened 2018-04-24
Start support Windows Containers managed by Kubernetes

xiaomi7732:
Address #105 and #115


src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
```diff
@@ -42,15 +43,28 @@ internal class K8sEnvironmentFactory : IK8sEnvironmentFactory
...
+                        containerId = myPod.Status.ContainerStatuses.First().ContainerID;
+                    }
+                    else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                    {
+                        // For Linux, container id could be fetched directly from cGroup.
+                        containerId = _httpClientSettings.ContainerId;
+                    }
+                    // ~
+
                     if (await SpinWaitContainerReady(timeoutAt, queryClient, containerId).ConfigureAwait(false))
```
karolz-ms: Is this still necessary given that you are waiting for the pod to be ready on line 47?

src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
```diff
@@ -42,15 +43,28 @@ internal class K8sEnvironmentFactory : IK8sEnvironmentFactory
...
                 using (K8sQueryClient queryClient = _k8sQueryClientFactory.Create(httpClient))
                 {
-                    string containerId = _httpClientSettings.ContainerId;
+                    // TODO: See if there's better way to fetch the container id
+                    K8sPod myPod = await SpinWaitUntilGetPod(timeoutAt, queryClient).ConfigureAwait(false);
+                    string containerId = null;
+                    if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                    {
+                        // For Windows, is there a way to fetch the current container id from within the container?
+                        containerId = myPod.Status.ContainerStatuses.First().ContainerID;
```
karolz-ms: Are you sure you will always be the first container in the pod?

karolz-ms: Also 
1. SpinWaitUntilGetPod() may return null
2. ContainerStatuses might be empty (low probability, but this is result of a network call, so sh*t happens)

src/ApplicationInsights.Kubernetes/K8sHttpClient/KubeHttpClientSettingsBase.cs
```diff
@@ -0,0 +1,145 @@
...
+                    return false;
+                }
+                else
+                {
+                    _logger.LogDebug(Invariant($"Issuer validation passed: {serverCertVerifier.Issuer}"));
+                }
+
+                // Server certificate is not expired.
+                DateTime now = DateTime.Now;
+                if (serverCertVerifier.NotBefore > now || serverCertVerifier.NotAfter.AddDays(1) <= now)
```
karolz-ms: Why 1 day of padding?

src/ApplicationInsights.Kubernetes/K8sHttpClient/KubeHttpClientSettingsBase.cs
```diff
@@ -0,0 +1,145 @@
...
+                {
+                    _logger.LogError(Invariant($"Server certification is not in valid period from {serverCertVerifier.NotBefore.ToString(DateTimeFormatInfo.InvariantInfo)} until {serverCertVerifier.NotAfter.ToString(DateTimeFormatInfo.InvariantInfo)}"));
+                    return false;
+                }
+                else
+                {
+                    _logger.LogDebug("Server certificate validate date verification passed.");
+                }
```
karolz-ms: Certificate validation should always check _three_ things:

1. Issued by trusted party
1. Not expired
1. Issued for service we are trying to communicate with

Seems like the third check is missing?

src/ApplicationInsights.Kubernetes/K8sHttpClient/KubeHttpSettingsWinContainerProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+            : base(kubernetesServiceHost, kubernetesServicePort, logger)
+        {
+            // Container id won't be fetched for windows container.
+            ContainerId = null;
+            DirectoryInfo serviceAccountDirectory =
+                new DirectoryInfo(Arguments.IsNotNullOrEmpty(serviceAccountFolder, nameof(serviceAccountFolder)));
+            foreach (FileSystemInfo fileSystemInfo in serviceAccountDirectory.EnumerateFileSystemInfos("*", SearchOption.AllDirectories))
+            {
+                // Per current symbolic linking settings, reading the file directly will fail.
+                if (!fileSystemInfo.Attributes.HasFlag(FileAttributes.ReparsePoint) && fileSystemInfo is FileInfo fileInfo)
```
karolz-ms: Nice, did not know this shortcut exists in C# now (`something is DesiredType desired`)

src/ApplicationInsights.Kubernetes/K8sHttpClient/KubeHttpSettingsWinContainerProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+            string namespaceFileName = "namespace",
+            string kubernetesServiceHost = null,
+            string kubernetesServicePort = null)
+            : base(kubernetesServiceHost, kubernetesServicePort, logger)
+        {
+            // Container id won't be fetched for windows container.
+            ContainerId = null;
+            DirectoryInfo serviceAccountDirectory =
+                new DirectoryInfo(Arguments.IsNotNullOrEmpty(serviceAccountFolder, nameof(serviceAccountFolder)));
+            foreach (FileSystemInfo fileSystemInfo in serviceAccountDirectory.EnumerateFileSystemInfos("*", SearchOption.AllDirectories))
```
karolz-ms: Since you are interested in files only, would `EnumerateFiles()` be sufficient?

karolz-ms: Actually, why enumerate at all? Can you just try testing for the 3 desired file names explicitly?

src/ApplicationInsights.Kubernetes/K8sQueryClient.cs
```diff
@@ -51,16 +56,30 @@ public Task<IEnumerable<K8sPod>> GetPodsAsync()
...
-                                pod.Status.ContainerStatuses != null &&
-                                pod.Status.ContainerStatuses.Any(
-                                    cs => !string.IsNullOrEmpty(cs.ContainerID) && cs.ContainerID.EndsWith(myContainerId, StringComparison.Ordinal)));
+
+            K8sPod targetPod = null;
+            string podName = Environment.GetEnvironmentVariable(@"APPINSIGHTS_KUBERNETES_POD_NAME");
+            string myContainerId = KubeHttpClient.Settings.ContainerId;
+            if (!string.IsNullOrEmpty(podName))
+            {
+                targetPod = allPods.FirstOrDefault(p => p.Metadata.Name.Equals(podName, StringComparison.Ordinal));
```
karolz-ms: Opportunity for improvement: if you know the pod name, use it in the query. In a large system there might be a lot of pods!

src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
xiaomi7732: They look similar but there is a minor difference. Line 47 is wait until the pod info is query-able. That is the earliest point where container id could be fetched; Here is waiting until the status of the pod become 'Ready'. There is a small gap between the 2 status.

src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
xiaomi7732: No, I am **not** sure. That is why I stated in the README.md that only **single container per pod** is supported for Windows container.
I **DO** want to review this later. If we could somehow get the container id from within the container itself, just like get it from cGroup in Linux, this problem will be solved automatically.

src/ApplicationInsights.Kubernetes/K8sHttpClient/KubeHttpClientSettingsBase.cs
xiaomi7732: Good catch. I can't remember the exact reason now. But here is my guess -
I thought the NotBefore / NotAfter is always aligned to 12:00am, which makes the valid case be:
NotBefore<=now<NotAfter+24hrs, then reverse the expression to see the invalid case, which is the code.
Well, problem is, NotBefore / NotAfter is not aligned to 12:00am, the valid case should be:
NotBefore <= now <= NotAfter. Reverse it should be:
NotBefore > now || NotAfter < now.

I'll fix it.

src/ApplicationInsights.Kubernetes/K8sHttpClient/KubeHttpSettingsWinContainerProvider.cs
xiaomi7732: Good catch. I'll update the code. Got distracted by the investigation on filtering out the symbolic link.

src/ApplicationInsights.Kubernetes/K8sQueryClient.cs
xiaomi7732: Good point. Filed a new issue: https://github.com/Microsoft/ApplicationInsights-Kubernetes/issues/114

src/ApplicationInsights.Kubernetes/K8sHttpClient/KubeHttpClientSettingsBase.cs
xiaomi7732: By current understand, the implementation of the certification validation is totally wrong.
**Details:**
By current understanding, it should be replaced as in https://github.com/kubernetes-client/csharp/blob/ea62ca00e099571be0e08022c27c44ef5f9392a5/src/KubernetesClient/Kubernetes.ConfigInit.cs
If possible, I would like to replace the whole query library with https://github.com/kubernetes-client/ when feasible.

src/ApplicationInsights.Kubernetes/K8sHttpClient/KubeHttpSettingsWinContainerProvider.cs
xiaomi7732: :-)

src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
karolz-ms: makes sense, thx for clarification

src/ApplicationInsights.Kubernetes/K8sHttpClient/KubeHttpClientSettingsBase.cs
xiaomi7732: Update: I guess I have a fix. Code will be updated later.

src/ApplicationInsights.Kubernetes/K8sHttpClient/KubeHttpSettingsWinContainerProvider.cs
karolz-ms: Roger

src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
david-jarman: The "COMPUTERNAME" environment variable in the container is set to a shortened version of the container id

xiaomi7732: @david-jarman From what I see, COMPUTERNAME is set to substring of the podName than the container id, is that not the case on your cluster?

examples/WindowsContainer/README.md
```diff
@@ -0,0 +1,69 @@
+At this stage (1.0.0-beta7+), we start to have limited support for Windows Container managed by Kubernetes. This example is intended to show you how to wire everything up. You are welcome to try this scenario. But please expect a slow response for any issue reported.
```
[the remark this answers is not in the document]
xiaomi7732: Rephrased.

xiaomi7732:
@david-jarman Thanks for the ask. Once we finish this PR, we will release a new beta package.
Could you please help us verify again with the latest change:
https://github.com/xiaomi7732/ApplicationInsights-Kubernetes/tree/dev/saars/win-container-1/examples/WindowsContainer
Please double check the package is:
```
<PackageReference Include="Microsoft.ApplicationInsights.Kubernetes" Version="1.0.0-private-201804252038" />
```

src/ApplicationInsights.Kubernetes/K8sHttpClient/KubeHttpClientSettingsBase.cs
```diff
@@ -0,0 +1,137 @@
...
+            // If the certificate is a valid, signed certificate, return true.
+            if (sslPolicyErrors == SslPolicyErrors.None)
+            {
+                _logger.LogTrace("This is a valid signed certificate.");
+                return true;
+            }
+
+            _logger.LogTrace("Not a authority signed certificate.");
+            _logger.LogTrace("Server Cert RAW: " + Environment.NewLine + Convert.ToBase64String(caCert.RawData));
+            // If there are errors in the certificate chain, look at each error to determine the cause.
```
karolz-ms: This is a comment taken from K8s client code, but it is misleading. The code below basically turns off certificate authority checking and certificate revocation checking, allowing certificates signed by unknown parties. In other words, as long as the certificate is not expired and it was issued for the right endpoint, it will be accepted regardless who issued it, or whether it was revoked.

Are you sure this is the right thing to do in this case?

If so, a better comment explaining the rationale would be very helpful here.

karolz-ms: If we are dealing with self-signed certificates here (self-signed by K8s cluster), maybe we can learn what the K8s cluster authority is and configure it explicitly as part of SSL checking policy (as a trusted authority). That would be **much better and stronger validation** than turning off cert authority check/cert revocation altogether.

src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
karolz-ms: SpinWaitUntilGetPod() may return null, meaning, on line 52 you will get NullRefException if it does

src/ApplicationInsights.Kubernetes/K8sHttpClient/KubeHttpClientSettingsBase.cs
xiaomi7732: :-) Yes, we are dealing with Self-Signed Certification here. My understand: the point of having cert validation is to prevent MITM. We are using the local cert + the returned server cert to build the chain. The chain must be valid. Refer line 101:
```csharp
      bool isValid = chain.Build(clientCertificate); // the server cert is in the chain object already
```

Unit tests has been added to verify both success & fail scenarios:
```
public void TrueWhenValidCertificate()
public void FalseWhenInvalidCertificate()
```

xiaomi7732: Comment updated.

xiaomi7732:
👍

david-jarman:
Bad news. I tried the latest package and the telemetry does not have the k8s info with it.

Every request to the kubernetes API server got a 403 Forbidden in response (189 calls in total):
```
dbug: Microsoft.ApplicationInsights.Kubernetes.K8sQueryClient[0]
      Query Failed. Request Message: Method: GET, RequestUri: 'https://10.0.0.1/api/v1/namespaces/default/pods', Version: 1.1, Content: <null>, Headers:
      {
        Authorization: Bearer <redacted>
      }. Status Code: Forbidden. Phase: Forbidden
dbug: Microsoft.ApplicationInsights.Kubernetes.K8sQueryClient[0]
```
At the end of the logs was this error:
```
dbug: Microsoft.ApplicationInsights.Kubernetes.K8sQueryClient[0]
      Query Failed. Request Message: Method: GET, RequestUri: 'https://10.0.0.1/api/v1/namespaces/default/pods', Version: 1.1, Content: <null>, Headers:
      {
        Authorization: Bearer <redacted>
        Request-Context: appId=cid-v1:d036e434-a5a1-499f-b6fb-94382dbf011d
        x-ms-request-root-id: e6ab8dd1-441698e3a2db6abc
        x-ms-request-id: |e6ab8dd1-441698e3a2db6abc.
        Request-Id: |e6ab8dd1-441698e3a2db6abc.
      }. Status Code: Forbidden. Phase: Forbidden
crit: Microsoft.ApplicationInsights.Kubernetes.K8sEnvironmentFactory[0]
      System.NullReferenceException: Object reference not set to an instance of an object.
         at Microsoft.ApplicationInsights.Kubernetes.K8sEnvironmentFactory.<CreateAsync>d__5.MoveNext()
fail: Microsoft.ApplicationInsights.Kubernetes.KubernetesTelemetryInitializer[0]
      Query Kubernetes Environment timeout.
```

Let me know if I can assist in any way.

Issue #121, opened 2018-05-26
Allow AppInsights.Kubernetes to run out of Kubernetes


xiaomi7732:
Allow this SDK to run in the environment that is not in Kubernetes. This enables one set of user code mixed environment.

PR #127, opened 2018-07-16
Dev/saars/no crash outside k8s

xiaomi7732:
Address #121


src/ApplicationInsights.Kubernetes/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -62,45 +58,59 @@ static ApplicationInsightsExtensions()
...
             return services;
         }
+
+        /// <summary>
+        /// Gets a logger for given type.
+        /// Note: This method leads to build service provider during the injection of services and shall be avoid whenever possible.
+        /// </summary>
+        private static ILogger<T> GetLogger<T>(this IServiceCollection services)
+        {
+            // According to the code, adding logging will not overwrite existing logging classes
```
karolz-ms: Just say that "`AddLogging` method is idempotent, safe to call multiple times on the same collection"

src/ApplicationInsights.Kubernetes/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -62,45 +58,59 @@ static ApplicationInsightsExtensions()
...
+            // https://github.com/aspnet/Logging/blob/c821494678a30c323174bea8056f43b93a3ca6f4/src/Microsoft.Extensions.Logging/LoggingServiceCollectionExtensions.cs
```
karolz-ms: Update reference so that it points to master branch, not some specific commit

src/ApplicationInsights.Kubernetes/Utilities/K8sDetectors/K8sDetector.cs
```diff
@@ -0,0 +1,26 @@
...
+        public K8sDetector(K8sDetector nextDetector)
+        {
+            _nextDetector = nextDetector;
+        }
+
+        /// <summary>
+        /// Returns true when the application is running in containers managed by Kubernetes.
+        /// </summary>
+        /// <returns></returns>
+        public bool IsRunningInKubernetes()
```
karolz-ms: This implementation is never used (every concrete class overrides it). Would a simple interface `IK8sDetector` be sufficient?

src/ApplicationInsights.Kubernetes/Debuggings/KubernetesDebuggingServiceCollectionBuilderFactory.cs
```diff
@@ -0,0 +1,23 @@
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.ApplicationInsights.Kubernetes.Debugging
+{
+    public sealed class KubernetesDebuggingServiceCollectionBuilderFactory
```
karolz-ms: Is this class being used? I do not see any reference to it, maybe I am missing something...

src/ApplicationInsights.Kubernetes/Utilities/K8sDetectors/K8sDetectorFactory.cs
```diff
@@ -0,0 +1,13 @@
+namespace Microsoft.ApplicationInsights.Kubernetes.Utilities
+{
+    internal sealed class K8sDetectorFactory
+    {
+        private K8sDetectorFactory() { }
+
+        public static K8sDetectorFactory Instance { get; } = new K8sDetectorFactory();
+
+#pragma warning disable CA1822 // Mark members as static
+        public K8sDetector CreateDetector() => new K8sDetectorLinux(new K8sDetectorWindows(null));
```
karolz-ms: I think this is a bit too clever for what it really does :) 

(I mean the detector chain)

src/ApplicationInsights.Kubernetes/Debuggings/KubernetesDebuggingServiceCollectionBuilderFactory.cs
xiaomi7732: No, it is not referenced. The intention of this class is to provide the end user a way to inject some fake data into their telemetry. This will only be used in the users' code when they prefer to.

I am not advertising this because I am not 100% sure the value for this. If no one asks (except you), I might just remove it later.

src/ApplicationInsights.Kubernetes/Utilities/K8sDetectors/K8sDetector.cs
xiaomi7732: As a matter of fact, this is where chain logic being considered, so that all the derived classes only need to consider its very specific prediction.
For example, the linux detector only detects for linux; the windows detector only detects for windows.

src/ApplicationInsights.Kubernetes/Extensions/ApplicationInsightsExtensions.cs
xiaomi7732: On github, when one clicks on the specific line and use 'Copy permalink', it will generate the path like that. My guess is that is the way to making sure the reference exists.

src/ApplicationInsights.Kubernetes/Utilities/K8sDetectors/K8sDetectorFactory.cs
xiaomi7732: Hmm, I don't really have a good reason to argue against the `too clever`. Do you really feel it becomes difficult to understand?

src/ApplicationInsights.Kubernetes/Extensions/ApplicationInsightsExtensions.cs
xiaomi7732: Sure. Firstly, learn a new word today - idempotent; secondly, the word itself seems a bit confused. Third, I'll simplify the comment in the next iteration.

src/ApplicationInsights.Kubernetes/Utilities/K8sDetectors/K8sDetector.cs
xiaomi7732: Off the topic, since you touched the topic on the interface, generally, I prefer to use an interface over an abstract class, simply because an interface is easier to mock. However, when there comes with shared logic, I think we shall not afraid of making use of it. What's your idea?

src/ApplicationInsights.Kubernetes/Extensions/ApplicationInsightsExtensions.cs
karolz-ms: https://stackoverflow.com/questions/1077412/what-is-an-idempotent-operation

src/ApplicationInsights.Kubernetes/Extensions/ApplicationInsightsExtensions.cs
karolz-ms: This is a link that refers to the version of the code as it was after a specific commit, namely, https://github.com/aspnet/Logging/commit/c821494678a30c323174bea8056f43b93a3ca6f4 

My point is, if you are making assumptions about how something is implemented, you should be willing to point to the latest version of the code, and not some implementation from the past, which might have changed since.

src/ApplicationInsights.Kubernetes/Utilities/K8sDetectors/K8sDetectorFactory.cs
karolz-ms: Well, yes, see my comment above, where I did not notice that there is a concrete implementation of a method that calls an abstract "Imp" method.

So you created an abstract base class with two methods, plus two concrete classes, for something that boils down to 

```csharp
() => Directory.Exists("here") || Directory.Exists("there")
```

src/ApplicationInsights.Kubernetes/Debuggings/KubernetesDebuggingServiceCollectionBuilderFactory.cs
karolz-ms: :/ 

I am not a big fan of adding features just in case someone *might* need it, but I will let you make the final call.

src/ApplicationInsights.Kubernetes/Utilities/K8sDetectors/K8sDetectorFactory.cs
xiaomi7732: :-) I guess in this instance, it will be a longshot for me to say that is for the extensibility (OC principle). It sometimes is really difficult to draw the line.

Despite this PR, how to do determine when to introduce some trick to make the code extensible vs when to write straightforward code and then expect no change happens?

src/ApplicationInsights.Kubernetes/Debuggings/KubernetesDebuggingServiceCollectionBuilderFactory.cs
xiaomi7732: More and more I am leaning toward to agree with you.
Not doing this for this PR since it is out of scope, I will seriously consider it.

src/ApplicationInsights.Kubernetes/Utilities/K8sDetectors/K8sDetectorFactory.cs
karolz-ms: Agree this is not always a straightforward call. I would consider the following:
  - How likely it is that we will have to introduce more flexible implementation in future?
  - If we have to change the implementation, how costly will the change be?

In this this case the i simplest interface is a `Func<Bool>`, the implementation is a one-liner, and it is not very likely to change in future, so I am inclined to go with the simple thing.

Issue #131, opened 2018-08-04
Build an images for troubleshooting


xiaomi7732:
Build an image with Application Insights Kubernetes enabled that runs when the K8s cluster is properly configured; And gather proper logs when it fails to run.

xiaomi7732:
- [ ] Image
- [ ] Instructions of how to use the image

PR #134, opened 2018-08-08
Add guidance of using troubleshooting image

xiaomi7732:
Addressing #131


troubleshooting/AIK8sTroubleShooting/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM microsoft/dotnet:2.1-sdk AS build-env
+WORKDIR /app
+
+# Copy csproj and restore as distinct layers
+COPY *.csproj ./
+RUN dotnet restore
+
+# Copy everything else and build
+COPY . ./
+RUN dotnet publish -c Release -o out
```
karolz-ms: Why not debug?

Just curious--it is for troubleshooting after all :)

troubleshooting/README.md
```diff
@@ -0,0 +1,63 @@
+# Troubleshooting the Environment
+Kubernetes cluster configuration could be complex. To run the app with `Application Insights for Kubernetes` on does not always work for the first shoot. It is a challenge when it is not clear where the issue comes from - the configuration for Kubernetes, the build of the images or the library of `Application Insights for Kubernetes`.
```
karolz-ms: .. does not always work on the first try

troubleshooting/AIK8sTroubleShooting/Dockerfile
xiaomi7732: I will keep it as is for now but I am totally open to change it to debug if necessary.
There is currently almost no user code for debugging per say. That might change later.

troubleshooting/README.md
xiaomi7732: Will fix this. Thanks!

PR #138, opened 2018-09-12
Update documentation EnableKubernetes

altenstedt:
This commit will update the documentation of
ApplicationInsightsExtensions.Enablekubernetes with information that
this method can be used to inject the Kubernetes telemetry initializer
if you are running in a non ASP.NET environment, like a console app
for example.


xiaomi7732:
Thanks for the contribution!

Issue #142, opened 2018-10-17
Update examples to reference the latest version of the packages


xiaomi7732:
For the sake of the security, the latest package is preferred. Update the examples to reference the latest packages.

PR #143, opened 2018-10-18
Update package versions

xiaomi7732:
Addresses #142 

1. Address vulnerability issues by bumping up external package versions in the examples.
2. Bump up the references to the NuGet package of AI.K8S itself to the latest non-prerelease version.


xiaomi7732:
Thanks for the quick turnaround!

Issue #145, opened 2018-10-31
Error "Value is required to set custom dimension. Key: Kubernetes.ReplicaSet.Name"


sarthakmahapatra:
Hi 

I upgraded Microsoft.ApplicationInsights.Kubernetes from 1.0.0-beta1 to 1.0.0. Started getting this error

``` c#
fail: Microsoft.ApplicationInsights.Kubernetes.KubernetesTelemetryInitializer[0]
      Value is required to set custom dimension. Key: Kubernetes.ReplicaSet.Name
```

this pod is running as a DeamonSet.

xiaomi7732:
@sarthakmahapatra Thanks for reporting the issue. DeamonSet is an interesting scenario. I'll take a look and then gets back to you.

sarthakmahapatra:
@xiaomi7732 thanks, looking forward to your response.

xiaomi7732:
@sarthakmahapatra The error is logged for debugging purpose. Although, I think should **not** be recorded as an error.
The PR will make the property optional so that the error message will only be seen in the debugging logging.

sarthakmahapatra:
Looks good, thanks

PR #147, opened 2018-11-06
Dev/saars/daemon set

xiaomi7732:
Address #145 
For pod created in DaemonSet, there's no ReplicaSet or Deployment available.
Since this info is not always available, they are made optional in this change.

### DIFF
--- a/src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
+++ b/src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
@@ -124,10 +124,10 @@ namespace Microsoft.ApplicationInsights.Kubernetes
 
             // Pod will have no replica name or deployment when deployed through other means. For example, as a daemonset.
             // Replica Set
-            SetCustomDimension(telemetry, Invariant($"{K8s}.{ReplicaSet}.Name"), this._k8sEnvironment.ReplicaSetName, true);
+            SetCustomDimension(telemetry, Invariant($"{K8s}.{ReplicaSet}.Name"), this._k8sEnvironment.ReplicaSetName, isValueOptional: true);
 
             // Deployment
-            SetCustomDimension(telemetry, Invariant($"{K8s}.{Deployment}.Name"), this._k8sEnvironment.DeploymentName, true);
+            SetCustomDimension(telemetry, Invariant($"{K8s}.{Deployment}.Name"), this._k8sEnvironment.DeploymentName, isValueOptional: true);
 
             // Ndoe
             SetCustomDimension(telemetry, Invariant($"{K8s}.{Node}.ID"), this._k8sEnvironment.NodeUid);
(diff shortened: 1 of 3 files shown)


src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
```diff
@@ -122,11 +122,12 @@ private void SetCustomDimensions(ITelemetry telemetry)
             SetCustomDimension(telemetry, Invariant($"{K8s}.{Pod}.Name"), this._k8sEnvironment.PodName);
             SetCustomDimension(telemetry, Invariant($"{K8s}.{Pod}.Labels"), this._k8sEnvironment.PodLabels);
 
+            // Pod will have no replica name or deployment when deployed through other means. For example, as a daemonset.
             // Replica Set
-            SetCustomDimension(telemetry, Invariant($"{K8s}.{ReplicaSet}.Name"), this._k8sEnvironment.ReplicaSetName);
+            SetCustomDimension(telemetry, Invariant($"{K8s}.{ReplicaSet}.Name"), this._k8sEnvironment.ReplicaSetName, true);
```
karolz-ms: Minor suggestion: for boolean flag parameter I like when the code is using named parameters, so it is immediately clear what `true` stands for, without looking up the method definition, e.g.

```csharp
SetCustomDimension(telemetry, Invariant($"{K8s}.{ReplicaSet}.Name"), this._k8sEnvironment.ReplicaSetName, isValueOptional: true);
```

src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
```diff
@@ -176,7 +177,14 @@ private void SetCustomDimension(ITelemetry telemetry, string key, string value)
 
             if (string.IsNullOrEmpty(value))
             {
-                _logger.LogError(Invariant($"Value is required to set custom dimension. Key: {key}"));
+                if (isValueOptional)
+                {
+                    _logger.LogTrace(Invariant($"Value is null or empty for key: {key}"));
```
karolz-ms: Do we even want to log this condition as a trace? It is *completely normal and expected* that a pod might not be created by a deployment, so there will be no deployment name. I am not sure if it helps with anything to have this in the logs.

src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
xiaomi7732: Sure.

src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
xiaomi7732: The trace level log reveals the code path, I don't see chatty becomes a problem. :-)

Issue #150, opened 2018-11-10
Refactor: split service injection with application running


xiaomi7732:
- [x] Split service injection with the execution
- [x] Take ILoggerFactory for extension method on ServiceCollotion

xiaomi7732:
Implemented with #151.

PR #151, opened 2018-11-17
Dev/saars/split service collection

xiaomi7732:
1. Address #150.
2. Add configuration support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,42 @@ We support **ASP.NET Core** application as well as **.NET Core** application.
 
 * Follow [this example](examples/BasicUsage_clr21_RBAC) for Role-based access control (RBAC) enabled Kubernetes clusters.
 
+### Configuration Details
+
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -24,6 +24,42 @@ We support **ASP.NET Core** application as well as **.NET Core** application.
...
+
+2. By `appsettings.json`:
+    ```jsonc
+    {
+        "Logging": {
+            // ...
+        },
+        // Adding the following section
+        "AppInsightsForKubernetes": {
+            "InitializationTimeout": "3.1:12:15.34"
```
karolz-ms: Is this 3+ days? Maybe a more practical value, like 15 seconds, would be more illustrative...

xiaomi7732: Shoot :-) You are right, it is a bad example in Readme. I'll update it. 👍

PR #176, opened 2019-01-31
[proposal] change cluster role name from metrics-reader to ai-k8s-role

ferantivero:
At first glance in some examples I couldn't find a correlation between the RBAC cluster role name ```metrics-reader``` and what I understand this role is actually being used for: _"fetch Kubernetes related properties"_ 

That being said, please let me propose changing its name to ```ai-k8s-role```.


karolz-ms:
@ferantivero thanks for your contribution!

It makes sense to change the name of the role as the current one is not very clear. However, I would suggest a different new name for the following reasons:

- The new name should convey the purpose of the role
- "ai" is most commonly used as an abbreviation for "artificial intelligence", not Application Insights

I propose `appinsights-k8s-property-reader` It is a bit verbose, but does convey the purpose well IMO. What do you think?

ferantivero:
that's definitely a better name -> addressed from https://github.com/Microsoft/ApplicationInsights-Kubernetes/commit/aded0a56420fe2fd55710b3b85253ebe6891cdc1

Issue #177, opened 2019-02-13
Enabling k8s support in a .Net Core app fails


brobichaud:
I'm trying to enable k8s support in my .Net Core app following the steps [here](https://github.com/Microsoft/ApplicationInsights-Kubernetes/wiki/Getting-Started-for-.NET-Core-Applications) and am having trouble.

It says to include the AspNetCore nuget, which I'm guessing is a copy/paste error from the equiv page for enabling it in AspNet apps? But even if I add that and the ApplicationInsights.Kubernetes nuget there is no EnableKubernetes() method to call. Clearly I'm doing something grossly wrong here. Note this is a Windows container, if that's relevant.

 Thoughts?

xiaomi7732:
@brobichaud Sorry for the confusion. Please follow the example of this for apps in Windows container: https://github.com/Microsoft/ApplicationInsights-Kubernetes/tree/develop/examples/WindowsContainer.

brobichaud:
But I am trying to enable it in a DotNet Core app, not an AspNet Core app. That link seems biased to AspNet Core.

xiaomi7732:
@brobichaud Please follow the Windows Container example above, but try using the following code to inject AI.K8s.
// I am sorry but I have limited internet access now that I can't verify it. Please post back if there's more problem:

```csharp
        static int Main(string[] args)
        {
            // Adding all environment variables into IConfiguration
            IConfiguration config = new ConfigurationBuilder().AddEnvironmentVariables().Build();

            // Read instrumentation key from IConfiguration
            string iKey = config["APPINSIGHTS_INSTRUMENTATIONKEY"];
            // Making sure ApplicationInsights instrumentation key is set
            if (string.IsNullOrEmpty(iKey))
            {
                Console.WriteLine("Please set the environment variable of APPINSIGHTS_INSTRUMENTATIONKEY with a property instrumentation key.");
                return -1;
            }

            // Build the client using the iKey to send trace.
            using (TelemetryConfiguration configuration = new TelemetryConfiguration(iKey))
            {
                configuration.AddApplicationInsightsKubernetesEnricher(applyOptions: null);

                TelemetryClient client = new TelemetryClient(configuration);
                while (true)
                {
                    // Send the trace
                    client.TrackTrace("Hello from AI SDK");
                    client.Flush();
                    Thread.Sleep(30000);
                }
            }
        }
```

ferantivero:
[...]

Issue #173, opened 2019-01-28
System.InvalidOperationException: A circular dependency was detected for the service of type 'Microsoft.Extensions.Logging.ILoggerFactory'


ferantivero:
I was trying to use the Hosting Startup way but I faced a circular reference issue with the new logging builder configuration for Application Insights. 

I was wondering whether there is a way to get both working together as of today:   

```
Unhandled Exception: System.InvalidOperationException: A circular dependency was detected for the service of type 'Microsoft.Extensions.Logging.ILoggerFactory'.
[...]

PR #178, opened 2019-03-05
Dev/saars/console example

xiaomi7732:
* Add an example for Console Application Insights to enable AI.K8s. Verify worked.
* Replace NuGet package deprecated licenseUrl property.
* Fix some typos in the code.
* Bump up the version number.


xiaomi7732:
As an exmaple for #177.

karolz-ms:
Sorry, I saw this and then is somehow fell through the cracks 😳 Looking now...

examples/BasicConsoleApp/README.md
```diff
@@ -0,0 +1,125 @@
+# Enable Application Insights for Kubernetes in .NET Core Console Application
+
+When running a Console Application inside a Kubernetes managed container, and if you have chosen Application Insights as your telemetry solution, you might want to enrich the telemetry with Kubernetes related information. This is an example of how to do it.
+
+_Note: This is an example that does not follow all best practices, including security-related best practices. E.g. Application Insights instrumentation key is not adequately protected (it should be deployed as a secret)._
+
+## Prerequisite
```
karolz-ms: prerequisites (plural)

src/.vscode/launch.json
```diff
@@ -0,0 +1,28 @@
+{
```
karolz-ms: VS Code is not required to run the sample so I would keep it minimal and not include this

src/.vscode/tasks.json
```diff
@@ -0,0 +1,57 @@
+{
```
karolz-ms: Remove?

xiaomi7732:
This example should address #173 too.

examples/BasicConsoleApp/README.md
xiaomi7732: Sure.

src/.vscode/launch.json
xiaomi7732: These are under src/ and are actually used to build the production code. They are relocated from the root of the repo.
The reason behind is to reduce the interference to the prod code from the examples.

I would not say this is the best place to put this change but at the same time, I don't think it worth another PR just for the relocate of these jsons.

xiaomi7732:
@karolz-ms  thanks for the review!

Issue #187, opened 2019-07-04
BasicConsoleAppILogger exemple not working


laulaua3:
Hello,
I try your exemple but I have no information of K8s in Application Insight:

2019-07-04T14:13:31.438 | Logger is working | 1 | trace | {"CategoryName":"XXXXX.Template.Program","{OriginalFormat}":"Logger is working","Method":"Main"} |   |   |   | PC | 0.0.0.0 | Dublin | Leinster | Ireland | testk8slog-7b66f99d6f-gxmrp.(none) | 69a7b2ed-4658-499f-acac-b24e670fcda5 | job-aks-app-insights | dbdffd18-4f83-4185-b480-c504e4309e86 | ai-k8s:1.1.0.0 | 23799490-9e55-11e9-b995-813b3150a92e | 1 |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
[...]

PR #188, opened 2019-07-10
Dev/saars/fix ilogger console example

xiaomi7732:
* Address #187 
* Update the code to send telemetry every 10 seconds.
* Provide examples for dockerization and K8s deployment.

### DIFF
--- a/examples/BasicConsoleAppILogger/Program.cs
+++ b/examples/BasicConsoleAppILogger/Program.cs
@@ -15,6 +15,14 @@ namespace BasicConsoleAppILogger
             // Create the DI container.
             IServiceCollection services = new ServiceCollection();
 
+            // Add application insights for Kubernetes. Making sure this is called before services.Configure<TelemetryConfiguration>().
+            services.AddApplicationInsightsKubernetesEnricher();
+
+            // Uncomment the following lines for debugging AI.K8s.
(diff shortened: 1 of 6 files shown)


xiaomi7732:
@karolz-ms Thanks for the comment. I'll get this in and keep iterating.

examples/BasicConsoleAppILogger/Program.cs
```diff
@@ -31,25 +41,28 @@ static void Main(string[] args)
                 // Application Insights for all categories.
                 builder.AddFilter<Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider>("", LogLevel.Trace);
                 builder.AddApplicationInsights("---Your AI instrumentation key---");
-            });
 
-            // Add application insights for Kubernetes.
-            services.AddApplicationInsightsKubernetesEnricher();
+                // Optional: Show the logs in console at the same time
+                builder.AddConsole();
```
GeorgeTsiokos: This call requires the `Microsoft.Extensions.Logging.Console` package, which is not referenced by the sample

xiaomi7732: @GeorgeTsiokos Thanks for pointing this out. Please file an issue and I'll fix it.

PR #192, opened 2019-07-19
do no return null from K8sQueryClient

bartosz6:
Right now, if you fail to fetch K8sEntityList, _null_ is returned. This leads to exception on application startup.

`System.ArgumentNullException: Value cannot be null.
Parameter name: source
   at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Func`2 predicate, Boolean& found)
   at Microsoft.ApplicationInsights.Kubernetes.K8sEnvironmentFactory.CreateAsync(DateTime timeoutAt)`

Btw. this message says absolutely nothing.

### DIFF
--- a/src/ApplicationInsights.Kubernetes/K8sQueryClient.cs
+++ b/src/ApplicationInsights.Kubernetes/K8sQueryClient.cs
@@ -55,12 +55,7 @@ namespace Microsoft.ApplicationInsights.Kubernetes
         {
             EnsureNotDisposed();
 
-            IEnumerable<K8sPod> allPods = await GetPodsAsync().ConfigureAwait(false);
-            if (allPods == null)
-            {
-                return null;
-            }
-
+            var allPods = await GetPodsAsync().ConfigureAwait(false);
             K8sPod targetPod = null;
             string podName = Environment.GetEnvironmentVariable(@"APPINSIGHTS_KUBERNETES_POD_NAME");
             string myContainerId = KubeHttpClient.Settings.ContainerId;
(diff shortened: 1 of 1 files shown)


xiaomi7732:
@bartosz6 Thanks for identifying the problem. The proposal will address the immediate crash, but it is incomplete because it will break the callers. Something like this: https://github.com/microsoft/ApplicationInsights-Kubernetes/blob/36245c926453d06cb82ed299445f04b3a017575d/src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs#L135

There is, unfortunately, a level of inconsistency.

To make the fix, there are 2 questions to be answer:
* What is the return value when the query fails, is that null or is that an enumerable.
* What the caller should expect, is that empty list or null when there's no element in the collection.

To follow the [best practice](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections), I think this PR is a step toward the right direction.

However, due to the existing implementation, all the callers needs to be updated as well.

@bartosz6 What do you think?

bartosz6:
> @bartosz6 Thanks for identifying the problem. The proposal will address the immediate crash, but it is incomplete because it will break the callers. Something like this:
> 
> https://github.com/microsoft/ApplicationInsights-Kubernetes/blob/36245c926453d06cb82ed299445f04b3a017575d/src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs#L135
> 
> There is, unfortunately, a level of inconsistency.
> 
> To make the fix, there are 2 questions to be answer:
> 
> * What is the return value when the query fails, is that null or is that an enumerable.
> * What the caller should expect, is that empty list or null when there's no element in the collection.
> 
> To follow the [best practice](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections), I think this PR is a step toward the right direction.
> 
> However, due to the existing implementation, all the callers needs to be updated as well.
> 
> @bartosz6 What do you think?

Agreed. I will check the dependent code and update the pull request.

bartosz6:
> See the comment in the conversation.

Turned out that there was only 1 dependent null check, that is now pointless because returned value is never null.

xiaomi7732:
@bartosz6 Good job! Thanks for the contribution!

Issue #203, opened 2020-04-10
The API path has been changed after k8s v1.16


hermanho:
The API path has been changed so the information of deployment and replicaset are empty.

Reference:
https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

xiaomi7732:
@hermanho, thanks for the heads up. I'll update related code.

xiaomi7732:
Released  in [1.1.2](https://www.nuget.org/packages/Microsoft.ApplicationInsights.Kubernetes/1.1.2).

PR #205, opened 2020-04-27
Dev/saars/add pod namespace

xiaomi7732:
* Add namespace to telemetry as Kubernetes.Pod.Namespace.
* Fixed a test certification expiration issue.

PR #209, opened 2020-05-01
Dev/saars/refresh query api

xiaomi7732:
Updated 2 query apis:
From | To
-- | --
apis/extensions/v1beta1/namespaces/{namespace}/replicasets | apis/apps/v1/namespaces/{namespace}/replicasets
apis/extensions/v1beta1/namespaces/{namespace}/deployments | apis/apps/v1/namespaces/{namespace}/deployments

Address #203


xiaomi7732:
Piggy backed a small bug fix: getting namespace from pod metadata object than pod object. #205

Issue #217, opened 2020-09-27
Disable Performance Counters


pinkfloydx33:
Is it possible to disable the two performance counter metrics (CPU and Memory) from being added as telemetry properties by the telemetry initializers? The CPU always returns `NaN` and the memory always reports some wildly inaccurate value (ex. 11gb when in actuality it's more like 200mb). These two statistics clutter all of my telemetry. Since we have other means of monitoring them anyways, they are just a bunch of noise that causes undue alarm (in the case of the Memory).

These two items aren't really Kubernetes related. I was honestly quite surprised when I finally tracked down where they were coming from to see that it was this library. Can we please get a way to toggle these unrelated properties so that we can omit them from telemetry?

xiaomi7732:
@pinkfloydx33 Thanks for the request. There's no way to configure the settings for now but I'll see if it can be added.

xiaomi7732:
Keep this open until the new version is released.

xiaomi7732:
Hi @pinkfloydx33 , the configuration is available at: 
https://www.nuget.org/packages/Microsoft.ApplicationInsights.Kubernetes/1.1.3

Refer to https://github.com/microsoft/ApplicationInsights-Kubernetes#configuration-details for how to use it.

pinkfloydx33:
@xiaomi7732 thank you, and thanks for the quick turn around!

PR #218, opened 2020-10-05
Dev/saars/config perf counter

xiaomi7732:
Address #217.

* Refactor the telemetries to separate the simple perf counter telemetry initializer with the primary k8s telemetry initializer
* Honor the user options to decide to include the simple perf counter telemetry initializer or not.
* Readme is updated with regarding the new option.
* CPU / Memory metrics have been moved to metrics of a telemetry entry from customer dimensions.


README.md
```diff
@@ -93,6 +93,7 @@ Customize configurations are supported starting with version 1.0.2 of the Applic
     | Key                   | Value/Types | Default Value | Description                                                                                  |
     | --------------------- | ----------- | ------------- | -------------------------------------------------------------------------------------------- |
     | InitializationTimeout | TimeSpan    | 00:02:00      | Maximum time to wait for spinning up the container. Accepted format: [d.]hh:mm:ss[.fffffff]. |
+    | DisableCounters       | Boolean     | false         | Sets to true to avoid adding performance counter telemetry initializer.                      |
```
karolz-ms, line 96: Consider `DisablePerformanceCounters` for extra clarity

src/ApplicationInsights.Kubernetes/TelemetryInitializers/SimplePerformanceCounterTelemetryInitializer.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            if (telemetry is ISupportMetrics metricsTelemetry)
+            {
+                // Write CPU metrics
+                TimeSpan endCPUTime = GetCPUTime();
+                if (_lastCPUSample != default)
+                {
+                    if (_cpuWatch.ElapsedMilliseconds > 0)
+                    {
+                        int processorCount = Environment.ProcessorCount;
```
karolz-ms, line 27: This does not seem to be used in the CPU time calculation ☺️ Also the value could probably be cached

src/ApplicationInsights.Kubernetes/TelemetryInitializers/SimplePerformanceCounterTelemetryInitializer.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        private void SetMetrics(ISupportMetrics telemetry, string key, double value)
+        {
+            if (!telemetry.Metrics.ContainsKey(key))
+            {
+                telemetry.Metrics.Add(key, value);
+            }
+        }
+
+        private TimeSpan GetCPUTime() => GetFromCurrentProcess<TimeSpan>(p => p.TotalProcessorTime);
```
karolz-ms, line 57: Creating and disposing of a `Process` object is somewhat expensive. Consider caching it.

karolz-ms (changes requested):

src/ApplicationInsights.Kubernetes/TelemetryInitializers/SimplePerformanceCounterTelemetryInitializer.cs
xiaomi7732: Good catch. It had been missed...

README.md
xiaomi7732: Point taken. Updated.

src/ApplicationInsights.Kubernetes/TelemetryInitializers/SimplePerformanceCounterTelemetryInitializer.cs
xiaomi7732: Fixed in the latest code.

src/ApplicationInsights.Kubernetes/TelemetryInitializers/SimplePerformanceCounterTelemetryInitializer.cs
xiaomi7732: Done.

src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
```diff
@@ -229,6 +193,6 @@ private static void SetCustomDimension(ISupportProperties telemetry, string key,
 
         internal bool _isK8sQueryTimeout = false;
         private bool _isK8sQueryTimeoutReported = false;
-        private static readonly ApplicationInsightsKubernetesDiagnosticSource _logger = ApplicationInsightsKubernetesDiagnosticSource.Instance;
+        protected static readonly ApplicationInsightsKubernetesDiagnosticSource _logger = ApplicationInsightsKubernetesDiagnosticSource.Instance;
```
xiaomi7732: Change back.

xiaomi7732: Done

src/ApplicationInsights.Kubernetes/TelemetryInitializers/SimplePerformanceCounterTelemetryInitializer.cs
```diff
@@ -0,0 +1,77 @@
...
+        private long GetMemory() => GetFromCurrentProcess<long>(p => p.VirtualMemorySize64);
+
+        private T GetFromCurrentProcess<T>(Func<Process, T> valueFactory)
+        {
+            // In case of concurrent execution, there is a chance for _currentProcess to be checked null more than one time,
+            // Take the hit to assign it multiple times to avoid locks.
+            if (_currnetProcess == null)
```
karolz-ms, line 69: ... or you can use `Lazy<T>` (just a thought)

src/ApplicationInsights.Kubernetes/TelemetryInitializers/SimplePerformanceCounterTelemetryInitializer.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace Microsoft.ApplicationInsights.Kubernetes
+{
+    internal sealed class SimplePerformanceCounterTelemetryInitializer : ITelemetryInitializer
+    {
+        private TimeSpan _lastCPUSample;
+        private Stopwatch _cpuWatch = new Stopwatch();
+
+        // Cache current process for performance trade off. 
+        // Refer to https://stackoverflow.com/questions/28023490/should-one-call-dispose-for-process-getcurrentprocess
```
karolz-ms, line 18: I think it is OK to not bother with disposing the Process object because `SimplePerformanceCounterTelemetryInitializer` is used as a **singleton** (one per Application Insights telemetry processor chain). That is a better justification IMO than relying on some Process implementation details.

karolz-ms (approved):
LGTM, just a couple minor comments.

src/ApplicationInsights.Kubernetes/TelemetryInitializers/SimplePerformanceCounterTelemetryInitializer.cs
xiaomi7732: Yes. Thanks for bring this up. Funny that I started with Lazy<T> and decide not to go with it just to make it even easier. The part that I am not 100% comfortable with Lazy<T> in this case is that Process is actually a IDisposable. I'll resolve the conversation for now but good point to think about.

src/ApplicationInsights.Kubernetes/TelemetryInitializers/SimplePerformanceCounterTelemetryInitializer.cs
xiaomi7732: Thanks for bring this up. I am sorry but I only partially agree :-).

I agree rely on the callee's implementation details is not ideal.

However, IMHO, it is as not ideal as rely on caller's implementation - there is NO constraint saying that `SimplePerformanceCounterTelemetryInitializer` or any `ITelemetryInitializer` have to be used as a singleton either.

That said, this crossed my mind, I just think it too heavy to implement for now. Post it here totally for discussion:
Basically, abstract out the current process part.
To create a singleton `CurrentProcessorInfoProvider`, that provides a method to get info from the current process. Implementation wise, current process object will be lazy load and cached, and it could implement IDisposable.

Let me know what do you think. :-D

karolz-ms: I like your proposal a lot, with just one caveat: as far as I can tell, the property getters of the `Process` class are not thread-safe, so the provider code will probably have to ensure that only one thread is accessing the `Process` instance at any given time

xiaomi7732: Task for another day. TBH, at this moment, I am more willing to rule out the SimplePerformanceCounterTelemetryInitializer. Here are my reasons:
1. **It is NOT k8s specific, and doesn't have to be in this repository;** (shall I say it shouldn't be here?)
1. .NET Core 3.x provides performance counters, I think it will be way more reliable to use that as the implementation to get metrics like CPU/Memory usage.

xiaomi7732: All that said, thank you so much for the review!

Issue #219, opened 2020-10-06
RBAC SARole suggestion


Olli64:
Greetings,

I stumbled across problem with multiple namespaces (dev, test, prod) with each having their own application insights.

As I'm not that familiar with kubernetes and dont have too much in-depth knowledge of the Roles/RoleBindings.

I had a small challenge here on how to make application insights able to collect information from all of the namespaces to enrich request telemetry.
(appins = application insights)
So I have 3 appins resources, ai-dev, ai-test and ai-prod.
I have multiple backend services with each having dev, test and prod environments (and linked to corresponding appins instance)
I have 3 namespaces in kubernetes (dev, test, prod)

Problem was that I could only get information from "prod" namespace by default and then I found this [github guide](https://github.com/microsoft/ApplicationInsights-Kubernetes/blob/develop/examples/BasicUsage_clr21_RBAC)

I added that clusterrolebinding to all namespaces just to notice that ofcourse it is overwriting it everytime because I was using same name (didnt understand that I could rename the ClusterRoleBinding to whatever)

Then finally I decided to create RoleBinding that referenced that ClusterRole to each namespace and finally after that I got it working.

Then at last I tried to rename those ClusterRoleBindings to different names in different namespaces and got it working also.

But what I couldn't figure out, why this is created as ClusterRoleBinding rather than RoleBinding?

As you must reference an service account from a namespace, that binds that ClusterRoleBinding to a namespace anyways (which loses the meaning of ClusterRoleBinding?)

So should it rather be RoleBinding which is scoped to a specific namespace as is the serviceaccount we should be giving the permissions to?

As in:
```
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: appinsights-k8s-property-reader
rules:
- apiGroups: ["","extensions", "apps"]
  resources: ["pods","nodes", "replicasets","deployments"]
  verbs: ["get", "watch", "list"]
---
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1beta1
metadata:
  name: appinsights-test-binding
subjects:
- kind: ServiceAccount
  name: default
  namespace: <test-namespace>
roleRef:
  kind: ClusterRole
  name: appinsights-k8s-property-reader
  apiGroup: rbac.authorization.k8s.io
---
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1beta1
metadata:
  name: appinsights-prod-binding
subjects:
[...]

PR #220, opened 2020-10-07
Dev/saars/update doc rbac - multiple namespaces

xiaomi7732:
Update the example due to #219


xiaomi7732:
@Olli64 thanks for the feedback. I updated the example accordingly.

karolz-ms (approved):
Looks great!

Issue #222, opened 2021-01-20
Current field naming enters in conflict with Kusto's data ingestion


feyanez:
Current naming convention for Kubernetes SDK's fields is not in compliance with the requirements by Kusto for data ingestion. In particular:
![image](https://user-images.githubusercontent.com/60495089/105251165-d8927700-5b2f-11eb-9d7d-a4992269550e.png)

For instance, in our case, the following fields are in conflict when Kusto attempts to create the tables since they were taken as "ID" and "Name" respectively. By having duplicate keys (i.e. different columns with same name), the table fails to be created.
![image](https://user-images.githubusercontent.com/60495089/105251400-4179ef00-5b30-11eb-92f3-aaa54b7a13ad.png)

As suggested from the Kusto team (see [IcM](https://portal.microsofticm.com/imp/v3/incidents/details/223704769/home) for more details), the owners of Kubernetes SDK should consider changing the naming of the fields to be underscore-separated instead of period-separated (e.g. "Kubernetes_Container_Name" instead of "Kubernetes.Container.Name"). Otherwise, it is needed to add the quotations as specified in the picture above.

Here is the [IcM](https://portal.microsofticm.com/imp/v3/incidents/details/221465008/home) on the Kubernetes SDK team for further reference.

After more than one month looking for the cause of the table-creation issue, I'm glad we are finally close to find a solution. Please let me know if you need any further information.

xiaomi7732:
To address the issue like it above, a meaningful way, for an SDK, is to allow customization of the telemetry names.

Consider adding options like this to allow pre-process for telemetry names:
```csharp
services.AddApplicationInsightsKubernetesEnricher(opt => {
    opt.TelemetryKeyProcesser = (key) => {
        return key.Replace('.', '_');  // Or whatever you want to do to reshape the key.
    }
});
```

xiaomi7732:
The feature is added in [1.1.4](https://www.nuget.org/packages/Microsoft.ApplicationInsights.Kubernetes/1.1.4). A wiki for the usage is here: [How to Customize Telemetry Property Names](https://github.com/microsoft/ApplicationInsights-Kubernetes/wiki/How-to-Customize-Telemetry-Property-Names).

PR #223, opened 2021-01-22
Add telemetry key cache to allow altering telemetry key scenario

xiaomi7732:
Addresses #222.

### DIFF
--- a/src/ApplicationInsights.Kubernetes/Extensions/AppInsightsForKubernetesOptions.cs
+++ b/src/ApplicationInsights.Kubernetes/Extensions/AppInsightsForKubernetesOptions.cs
@@ -21,5 +21,11 @@ namespace Microsoft.Extensions.DependencyInjection
         /// </summary>
         [JsonProperty("DisablePerformanceCounters")]
         public bool DisablePerformanceCounters { get; set; }
+
+        /// <summary>
+        /// Gets or sets the processor for telemetry key. This is introduced to allow customization of
+        /// telemetry keys.
+        /// </summary>
+        public Func<string, string> TelemetryKeyProcessor { get; set; }
     }
 }
--- a/src/ApplicationInsights.Kubernetes/Extensions/KubernetesServiceCollectionBuilder.cs
+++ b/src/ApplicationInsights.Kubernetes/Extensions/KubernetesServiceCollectionBuilder.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Extensions.DependencyInjection
                 InjectChangableServices(serviceCollection);
 
                 serviceCollection.AddSingleton<ITelemetryInitializer, KubernetesTelemetryInitializer>();
-
+                
 #if NETSTANDARD2_0
                 if (_options.Value == null || !_options.Value.DisablePerformanceCounters)
                 {
@@ -71,6 +71,7 @@ namespace Microsoft.Extensions.DependencyInjection
 
         private static void InjectCommonServices(IServiceCollection serviceCollection)
         {
+            serviceCollection.AddSingleton<ITelemetryKeyCache, TelemetryKeyCache>();
             serviceCollection.AddSingleton<KubeHttpClientFactory>();
             serviceCollection.AddSingleton<K8sQueryClientFactory>();
             serviceCollection.AddSingleton<SDKVersionUtils>(SDKVersionUtils.Instance);
--- a/src/ApplicationInsights.Kubernetes/Extensions/KubernetesServiceCollectionBuilder.cs
+++ b/src/ApplicationInsights.Kubernetes/Extensions/KubernetesServiceCollectionBuilder.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Extensions.DependencyInjection
                 InjectChangableServices(serviceCollection);
 
                 serviceCollection.AddSingleton<ITelemetryInitializer, KubernetesTelemetryInitializer>();
-                
+
 #if NETSTANDARD2_0
                 if (_options.Value == null || !_options.Value.DisablePerformanceCounters)
                 {
--- a/src/ApplicationInsights.Kubernetes/TelemetryInitializers/ITelemetryKeyCache.cs
+++ b/src/ApplicationInsights.Kubernetes/TelemetryInitializers/ITelemetryKeyCache.cs
@@ -2,6 +2,6 @@ namespace Microsoft.ApplicationInsights.Kubernetes
 {
     internal interface ITelemetryKeyCache
     {
-        string GetTelemetryProcessedKey(string originalKey);
+        string GetProcessedKey(string originalKey);
     }
 }
--- a/src/ApplicationInsights.Kubernetes/TelemetryInitializers/ITelemetryKeyCache.cs
+++ b/src/ApplicationInsights.Kubernetes/TelemetryInitializers/ITelemetryKeyCache.cs
@@ -1,0 +1,7 @@
+namespace Microsoft.ApplicationInsights.Kubernetes
+{
+    internal interface ITelemetryKeyCache
+    {
+        string GetTelemetryProcessedKey(string originalKey);
+    }
+}
--- a/src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
+++ b/src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
@@ -153,7 +153,7 @@ namespace Microsoft.ApplicationInsights.Kubernetes
 
         private void SetCustomDimension(ISupportProperties telemetry, string key, string value, bool isValueOptional = false)
         {
-            key = _telemetryKeyCache.GetTelemetryProcessedKey(key);
+            key = _telemetryKeyCache.GetProcessedKey(key);
 
             if (telemetry == null)
             {
--- a/src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
+++ b/src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
@@ -19,18 +19,17 @@ namespace Microsoft.ApplicationInsights.Kubernetes
     /// </summary>
     internal class KubernetesTelemetryInitializer : ITelemetryInitializer
     {
+        private static readonly ApplicationInsightsKubernetesDiagnosticSource _logger = ApplicationInsightsKubernetesDiagnosticSource.Instance;
+        
         private readonly SDKVersionUtils _sdkVersionUtils;
         private readonly DateTime _timeoutAt;
-
         internal readonly IK8sEnvironmentFactory _k8sEnvFactory;
+        internal readonly AppInsightsForKubernetesOptions _options;
+        internal readonly ITelemetryKeyCache _telemetryKeyCache;
         internal IK8sEnvironment _k8sEnvironment { get; private set; }
-
-        internal AppInsightsForKubernetesOptions _options { get; private set; }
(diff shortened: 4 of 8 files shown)


src/ApplicationInsights.Kubernetes/Extensions/KubernetesServiceCollectionBuilder.cs
```diff
@@ -47,7 +47,7 @@ public IServiceCollection InjectServices(IServiceCollection serviceCollection)
                 InjectChangableServices(serviceCollection);
 
                 serviceCollection.AddSingleton<ITelemetryInitializer, KubernetesTelemetryInitializer>();
-
+
```
xiaomi7732, line 50: Remove

src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
```diff
@@ -19,10 +19,24 @@ namespace Microsoft.ApplicationInsights.Kubernetes
     /// </summary>
     internal class KubernetesTelemetryInitializer : ITelemetryInitializer
     {
+        private readonly SDKVersionUtils _sdkVersionUtils;
+        private readonly DateTime _timeoutAt;
+
+        internal readonly IK8sEnvironmentFactory _k8sEnvFactory;
+        internal IK8sEnvironment _k8sEnvironment { get; private set; }
```
xiaomi7732, line 26: Are these private setters needed?

tests/UnitTests/TelemetryKeyCacheTests.cs
```diff
@@ -0,0 +1,59 @@
...
+            IOptions<AppInsightsForKubernetesOptions> options = Options.Create<AppInsightsForKubernetesOptions>(opts);
+            TelemetryKeyCache keyCache = new TelemetryKeyCache(options);
+
+            string actual = keyCache.GetTelemetryProcessedKey(input);
+
+            Assert.Equal(expected, actual);
+        }
+
+        [Fact]
+        public void ReturnsOriginalWhenProcessorNotSet()
```
feyanez, line 43: Could you add a test that succeeds on parsing "Abc.Def" to "AbcDef"?

xiaomi7732: That would be easy to add. Mind explain me why?

src/ApplicationInsights.Kubernetes/Extensions/AppInsightsForKubernetesOptions.cs
```diff
@@ -21,5 +21,11 @@ public class AppInsightsForKubernetesOptions
         /// </summary>
         [JsonProperty("DisablePerformanceCounters")]
         public bool DisablePerformanceCounters { get; set; }
+
+        /// <summary>
+        /// Gets or sets the processor for telemetry key. This is introduced to allow customization of
+        /// telemetry keys.
+        /// </summary>
+        public Func<string, string> TelemetryKeyProcessor { get; set; }
```
karolz-ms, line 29: Alternatively this could be a setting, or a couple of settings: a search pattern, and a replacement string, using either [ordinary string replacement](https://docs.microsoft.com/en-us/dotnet/api/system.string.replace?view=net-5.0#System_String_Replace_System_String_System_String_System_StringComparison_) or [regex replace](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.replace?view=net-5.0#System_Text_RegularExpressions_Regex_Replace_System_String_System_String_System_String_System_Text_RegularExpressions_RegexOptions_System_TimeSpan_). 

Regex can be costly, so I would do it as opt-in, but it really depends on the specific regular expression, and we would probably compile the regexp for fast execution.

src/ApplicationInsights.Kubernetes/TelemetryInitializers/ITelemetryKeyCache.cs
```diff
@@ -0,0 +1,7 @@
+namespace Microsoft.ApplicationInsights.Kubernetes
+{
+    internal interface ITelemetryKeyCache
+    {
+        string GetTelemetryProcessedKey(string originalKey);
```
karolz-ms, line 5: Nit: `GetProcessedKey()` or `GetProcessedTelemetryKey()`?

src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
```diff
@@ -33,6 +46,7 @@ internal class KubernetesTelemetryInitializer : ITelemetryInitializer
             _logger.LogDebug(@"Initialize Application Insights for Kubernetes telemetry initializer with Options:
 {0}", JsonConvert.SerializeObject(_options));
 
+            _telemetryKeyCache = telemetryKeyCache ?? throw new ArgumentNullException(nameof(telemetryKeyCache));
```
karolz-ms, line 49: Or use an instance of the (standard) `TelemetryCache` if nothing is passed to the constructor

src/ApplicationInsights.Kubernetes/Extensions/AppInsightsForKubernetesOptions.cs
karolz-ms: Also, this contract does not prevent the telemetry key processor to return **different** transformed key for the same input (upon every invocation). Maybe we consider this a "misbehaving" processor, but we should probably document this, otherwise the caching of the transform result is not possible.

src/ApplicationInsights.Kubernetes/TelemetryInitializers/TelemetryKeyCache.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public string GetTelemetryProcessedKey(string originalKey)
+        {
+            // Implicit condition: telemetry cache stays null means the key processor is not set in options.
+            if (_telemetryKeyCache == null)
+            {
+                return originalKey;
+            }
+
+            return _telemetryKeyCache.GetOrAdd(originalKey, _options.TelemetryKeyProcessor.Invoke(originalKey));
```
karolz-ms, line 31: There is no limit on the key cache growth...

karolz-ms (commented):
No objections, but some questions ☺️

tests/UnitTests/TelemetryKeyCacheTests.cs
feyanez: I just think that some team's conventions could be replacing Kubernetes.Container.Name for KubernetesContainerName, for example. So just want to make sure the test for that scenario passes to assure it is a possibility.

src/ApplicationInsights.Kubernetes/TelemetryInitializers/ITelemetryKeyCache.cs
xiaomi7732: `GetProcessedKey()` makes sense. I'll update.

src/ApplicationInsights.Kubernetes/TelemetryInitializers/TelemetryKeyCache.cs
xiaomi7732: Being lazy, I think I'll just put a cap on the dictionary.

src/ApplicationInsights.Kubernetes/Extensions/AppInsightsForKubernetesOptions.cs
xiaomi7732: This is a great question. It is tough to balance functionality and security. Being hit off guard by this requirement makes me want to provide a way to allow address all cases. For example, the user could prefer to hash the key rather than simple replacement; That said, I see your concerns on ever growing transformed-key list.

I think I'll `document it` + `limit the cap of the cache entry number` for now.

src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
xiaomi7732: Are you expecting something like this?
```csharp
_telemetryKeyCache = telemetryKeyCache ?? TelemetryKeyCache.Default;
```

If that is the case: no. :-) In sprite of making the dependency explicit, ITelemetryKeyCache is required, not nullable.

Thinking on this is: I wouldn't mind to provide an implementation of `DefaultTelemetryKeyCache` if that interface is exposed (public). But since this is a requirement for our own code (internals), I prefer to keep it this way.

Feel free to start up a discussion offline with me :-)

src/ApplicationInsights.Kubernetes/TelemetryInitializers/TelemetryKeyCache.cs
```diff
@@ -0,0 +1,49 @@
...
+        private ConcurrentDictionary<string, string> _telemetryKeyCache;
+
+        public TelemetryKeyCache(IOptions<AppInsightsForKubernetesOptions> options)
+        {
+            _options = options?.Value ?? throw new System.ArgumentNullException(nameof(options));
+
+            if (_options.TelemetryKeyProcessor != null)
+            {
+                _telemetryKeyCache = new ConcurrentDictionary<string, string>(
+                    concurrencyLevel: 16,
```
karolz-ms, line 25: Probably not necessary. According to https://devblogs.microsoft.com/pfxteam/concurrentdictionary-performance-improvements-in-net-4-5/ the ConcurrentDictionary will dynamically change the number of locks used if concurrencyLevel is not specified.

src/ApplicationInsights.Kubernetes/TelemetryInitializers/TelemetryKeyCache.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.ApplicationInsights.Kubernetes
+{
+    internal class TelemetryKeyCache : ITelemetryKeyCache
+    {
+        // The capacity of 40 is kind of arbitrary. We have around 10 keys appended to for the telemetry;
+        // Times that by 4 to allow a bit of wiggling room for name mapping;
+        // However, having too many of the mapped key is not allowed to prevent huge memory consuption.
+        internal const int CacheCapacity = 40;
```
karolz-ms, line 13: I did not realize the number of keys that will be held by the cache is fixed, sorry. If that is the case though, then there is no concern of unlimited growth and the fixed capacity is not necessary IMO.

src/ApplicationInsights.Kubernetes/TelemetryInitializers/TelemetryKeyCache.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public string GetProcessedKey(string originalKey)
+        {
+            // Implicit condition: telemetry cache stays null means the key processor is not set in options.
+            if (_telemetryKeyCache == null)
+            {
+                return originalKey;
+            }
+
+            string result = _telemetryKeyCache.GetOrAdd(originalKey, _options.TelemetryKeyProcessor.Invoke(originalKey));
```
karolz-ms, line 39: You should use `GetOrAdd()` overload that takes a value factory function. Do you see why this is very important? ☺️

karolz-ms (changes requested):

src/ApplicationInsights.Kubernetes/TelemetryInitializers/TelemetryKeyCache.cs
xiaomi7732: Ooops. Fixed.

karolz-ms (approved):
🚀

PR #225, opened 2021-04-01
Add an example

xiaomi7732:
* .NET 5
* Using Application Insights SDK with connection string
* Docker: Simplified dockerfile to build the binary locally
* K8s: Example of using NodePort for service to debug locally

### DIFF
--- a/examples/BuildOutOfContainer/BuildOutOfContainer.csproj
+++ b/examples/BuildOutOfContainer/BuildOutOfContainer.csproj
@@ -1,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.17.0" />
+    <PackageReference Include="Swashbuckle.AspNetCore" Version="5.6.3" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\ApplicationInsights.Kubernetes\ApplicationInsights.Kubernetes.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/examples/BuildOutOfContainer/Readme.md
+++ b/examples/BuildOutOfContainer/Readme.md
@@ -1,0 +1,66 @@
+# Example of building the application out of container
+
+This is an example when building the application out of the container.
+
+## Build the container
+
+* Build/Publish the binaries locally
+
+    ```shell
+    dotnet publish -c Release
+    ```
+
+* If it is the first time running, pull the base image:
+
+    ```shell
+    docker pull mcr.microsoft.com/dotnet/aspnet:5.0
+    ```
+
+* Build the container:
+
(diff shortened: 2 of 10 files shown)


examples/BuildOutOfContainer/BuildOutOfContainer.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <TargetFramework>net5.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.17.0" />
+    <PackageReference Include="Swashbuckle.AspNetCore" Version="5.6.3" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\ApplicationInsights.Kubernetes\ApplicationInsights.Kubernetes.csproj" />
```
karolz-ms, line 13: Consider forward slash if you want the project to be usable on Mac/Linux

examples/BuildOutOfContainer/Startup.cs
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        public IConfiguration Configuration { get; }
+
+        // This method gets called by the runtime. Use this method to add services to the container.
+        public void ConfigureServices(IServiceCollection services)
+        {
+            // Application Insights connection string is set in appsettings.json
+            services.AddApplicationInsightsTelemetry();
+            // Skip the following lines when it is in Production environment
```
karolz-ms, line 26: Is there an equivalent of `env.IsDevelopment()` that you could use here?

examples/BuildOutOfContainer/Readme.md
```diff
@@ -0,0 +1,66 @@
...
+
+```shell
+kubectl delete -f .\K8s.yaml
+```
+
+Got:
+> deployment.apps "ai-k8s-webapi" deleted  
+service "ai-k8s-oc-webapi-svc" deleted  
+
+
```
karolz-ms, line 65: What about app settings/App Insights configuration, do you want to cover that too?

karolz-ms (approved):
LGTM, just a few questions/suggestions!

examples/BuildOutOfContainer/BuildOutOfContainer.csproj
xiaomi7732: Having a little bit faith in `csc` :-) It understands the path for all OSes :-) Verified it on Ubuntu. Probably won't fix this. Either unfair to Windows Users or unfair to Unix users :-)

examples/BuildOutOfContainer/Startup.cs
xiaomi7732: The way I can think of is injecting `IWebHostEnvironment`. Will update.

examples/BuildOutOfContainer/Readme.md
xiaomi7732: Sure. Why not. :-) Updated the doc. Thanks for the feedback.

examples/BuildOutOfContainer/Startup.cs
xiaomi7732: Fixed.

PR #233, opened 2021-06-15
Update Newtonsoft.Json naar 13.0.1 to mitigate DOS vulnerability

304NotModified:
Because of https://github.com/JamesNK/Newtonsoft.Json/issues/2535 and https://alephsecurity.com/vulns/aleph-2018004


xiaomi7732 (approved):
Looks good to me. Thanks for your contribution.

304NotModified:
@xiaomi7732 

Thanks for the merge. Any idea when the release is scheduled? (and which version number?)

304NotModified:
AH i see the beta. I not sure if 2.1 would be better according to semver

xiaomi7732:
There's no functional/api change but straight re-compile. It is treated as a patch update (for the re-compile). 2.0.2 will be the target version from 2.0.1 when it is stable.

Issue #236, opened 2021-08-04
AddApplicationInsightsKubernetesEnricher breaks ASP.NET Core IConfiguration reload notification system.


jimmyca15:
ASP.NET Core has a way to monitor configuration changes for the default IConfiguration registered in DI.

E.g.

```
            IConfiguration config = app.ApplicationServices.GetRequiredService<IConfiguration>();

            ChangeToken.OnChange(config.GetReloadToken,
                () => {
                    Console.WriteLine("abc");
                });
```

Calling `services.AddApplicationInsightsKubernetesEnricher` in an ASP.NET Core application causes this reload notification to break.

Here is a simple [repro application](https://github.com/jimmyca15/k8ai-reloadrepro/blob/845c062019715c748e5fe1871ece66e410efd434/ReloadRepro/Startup.cs#L34).  

To use the repro, start the application and update the appsettings.json file while it is running. "abc" will output to the terminal.

To repro the issue, uncomment [this line](https://github.com/jimmyca15/k8ai-reloadrepro/blob/845c062019715c748e5fe1871ece66e410efd434/ReloadRepro/Startup.cs#L34) `//services.AddApplicationInsightsKubernetesEnricher();    // Enables Application Insights for Kubernetes.
        }`. Then run the application and update the appsettings.json file. Nothing will be output to the terminal because the configuration reload token will never be triggered.

xiaomi7732:
Hi @jimmyca15 thanks for bring this up. I'll take a look.

xiaomi7732:
@jimmyca15, checkout the new package with the fix:
https://www.nuget.org/packages/Microsoft.ApplicationInsights.Kubernetes/2.0.2-beta2

Thanks!

emanuelpaul:
> @jimmyca15, checkout the new package with the fix: https://www.nuget.org/packages/Microsoft.ApplicationInsights.Kubernetes/2.0.2-beta2
> 
> Thanks!

@xiaomi7732 is it possible to create a release version ? it seems to work fine but it's a bit confusing that only pre-release package exists with this fix.

xiaomi7732:
Totally possible. I'll make that happen ASAP.

emanuelpaul:
Thank you !

PR #237, opened 2021-08-29
Avoid building pre-mature service provider

xiaomi7732:
Address #236.

The issue is caused by the fact a pre-matured service provider is built during the service registration. Bad practice:
Code:
https://github.com/microsoft/ApplicationInsights-Kubernetes/blob/ad6974f1c994c3f48e6ce11e779b878a7203c92a/src/ApplicationInsights.Kubernetes/Extensions/ApplicationInsightsExtensions.cs#L127


xiaomi7732:
Tag  @jimmyca15 who reported the issue for awareness.

karolz-ms (approved):
LGTM ☺️

Issue #253, opened 2022-05-18
Kubernetes related properties are not coming up in App insights


srivathsanvlb:
I am doing the below steps.

1. I have created a simple .NET 6 console app as below

```
namespace LoggingPOC.K8s
{
    public class Program
    {
        public static void Main(string[] args)
        {
            IServiceCollection services = new ServiceCollection();         

            services.AddApplicationInsightsKubernetesEnricher();

            var channel = new InMemoryChannel();
            services.Configure<TelemetryConfiguration>(
                (config) =>
                {
                    config.TelemetryChannel = channel;
                }
            );
            services.AddLogging(builder =>
            {                
                builder.AddFilter<Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider>("", LogLevel.Trace);
                builder.AddApplicationInsights(MyKey);
                builder.AddConsole();
            });

            IServiceProvider serviceProvider = services.BuildServiceProvider();
            ILogger<Program> logger = serviceProvider.GetRequiredService<ILogger<Program>>();

            while (true)
            {
                using (logger.BeginScope(new Dictionary<string, object> { { "Method", nameof(Main) } }))
                {
                    logger.LogInformation("Logger is working"); // this will be captured by Application Insights.
                }
                channel.Flush();
                Thread.Sleep(10000);
            }
        }
    }
}
```

2. I have dockerized the above app and have placed it in a AKS cluster. 

3. The logs that I have written are appearing correctly. But no Kubernetes related information similar to [this](https://github.com/microsoft/ApplicationInsights-Kubernetes/blob/develop/docs/TelemetryEnhancement.png) is showing up.

<img width="323" alt="Untitled" src="https://user-images.githubusercontent.com/35536827/168987298-d3c44cce-6fe1-4b5f-a19b-df0ea869c747.png">

Kindly assist me in what I am missing 

```
		<PackageReference Include="Microsoft.ApplicationInsights.Kubernetes" Version="2.0.3" />
		<PackageReference Include="Microsoft.Extensions.Hosting" Version="6.0.1" />
		<PackageReference Include="Microsoft.Extensions.Logging.ApplicationInsights" Version="2.20.0" />
		<PackageReference Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="1.14.0" />

		<PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="6.0.1" />
[...]

PR #260, opened 2022-05-20
Dev/saars/update console example

xiaomi7732:
Update example code due to application insights behavior change. See details in #253 .


karolz-ms (approved):
Looks good. Sorry for late reply; I was OOF on Friday.

xiaomi7732:
@karolz-ms Thanks for the review. You have nothing to say sorry for. I think, a PR submitted on Friday, reviewed by Monday morning is a very quick turn-around.
Thank you for always being there.

PR #263, opened 2022-05-21
Dev/saars/compliance docker registry

xiaomi7732:
The primary goal is to update examples / troubleshooting container example files to use MCR. This is a compliance requirement.


karolz-ms (approved):
Looks good, just a couple of questions/suggestions

Issue #266, opened 2022-05-23
Unit test is failing


xiaomi7732:
Due to #264 , unit test failed because of the signature change.

* Also look into why we don't have a PR build before merging.

xiaomi7732:
Unit test fixed. GitHub action is set up. PR #268 for tracking. Close this one.

PR #268, opened 2022-05-23
Add CI to Github actions

xiaomi7732:
Part of #266. Unfortunately, I don't have a quick way to verify github actions before merging the change.


xiaomi7732:
It looks like this would have caught the unit tests failing issue. Waiting for #267 to be merged first.

xiaomi7732:
Interesting, the same unit test passed in Azure Pipeline but failed in github action CI build...

xiaomi7732:
It looks like switching to windows-latest worked the problem around. It is weird because I couldn't repo the same issue on a local box of Ubuntu 22.04.

xiaomi7732:
Also passes on ubuntu-22.04 in github action. I suspect there is an issue for the code specifically on 20.04.

karolz-ms (approved):

Issue #246, opened 2021-12-22
Not compatible with cgroupv2


pkoelbl:
This is not compatible with cgroupv2 and causes an error when initializing Application Insights Kubernetes
KubeHttpClientSettingsProvider.cs line 25 string pathToCGroup = @"/proc/self/cgroup" 

See this for details
https://stackoverflow.com/questions/68816329/how-to-get-docker-container-id-from-within-the-container-with-cgroup-v2

xiaomi7732:
Hi @pkoelbl Thanks for the reporting.

I tried it on docker 20.10.10 with the latest alpine and it appears still working:

![image](https://user-images.githubusercontent.com/3674549/150605969-75ef222f-f3d9-4512-82fd-c3c657797787.png)

Is there anything else needed to reproduce the issue? Thanks!

pkoelbl:
I am on a Mac using Docker Desktop 4.4.2(73395) Docker engine 20.10.12 with Kubernetes 1.22.5 enabled. Using kubernetes to create the pod. Using mcr.microsoft.com/dotnet/aspnet:5.0-buster-slim as the base image. When i connect to the pod and look at the cgroup file it is empty.

pkoelbl:
The app is also running as a non-root user.

xiaomi7732:
@pkoelbl Thanks for the update. We will look into that.

xiaomi7732:
@pkoelbl I updated the env (Docker Desktop: 4.5.1, Engine: 20.10.12, K8s:1.22.5), am still able to access cgroups:

```shell
PS D:\> docker run -it mcr.microsoft.com/dotnet/aspnet:5.0-buster-slim
root@ca7739f0a9f1:/# cat /proc/self/cgroup
14:name=systemd:/docker/ca7739f0a9f1ab77959378e36b48c22d68410adae6db2e44be7b04271e49a81c
13:rdma:/
12:pids:/docker/ca7739f0a9f1ab77959378e36b48c22d68410adae6db2e44be7b04271e49a81c
11:hugetlb:/docker/ca7739f0a9f1ab77959378e36b48c22d68410adae6db2e44be7b04271e49a81c
10:net_prio:/docker/ca7739f0a9f1ab77959378e36b48c22d68410adae6db2e44be7b04271e49a81c
9:perf_event:/docker/ca7739f0a9f1ab77959378e36b48c22d68410adae6db2e44be7b04271e49a81c
[...]

PR #273, opened 2022-05-27
Added a container id provider that supports getting info from mountinfo

xiaomi7732:
Addresses #246 

Looks like `/proc/self/mountinfo` is a relative reliable spot to getting container id that supports cgroupv2.

* Piggy back on it, I also bumped up the language used in the project to C# 10.
* A small improvement on logging.


xiaomi7732:
@pkoelbl This is the PR to implement getting container id from mountinfo. Let me know if you are willing to share any early feedback.

pkoelbl:
Looks good. Let me know when you have a beta nuget

src/ApplicationInsights.Kubernetes/ContainerIdProviders/MountInfoContainerIdProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Text.RegularExpressions;
+using Microsoft.ApplicationInsights.Kubernetes.Debugging;
+
+namespace Microsoft.ApplicationInsights.Kubernetes.ContainerIdProviders;
+
+internal class MountInfoContainerIdProvider : IContainerIdProvider
+{
+    private const string InfoFilePath = "/proc/self/mountinfo";
+    private const string MatchPattern = @"\/docker\/containers\/(.*?)\/";
+    private static readonly Regex MatchRegex = new Regex(MatchPattern, RegexOptions.CultureInvariant | RegexOptions.Multiline, TimeSpan.FromSeconds(1));
```
karolz-ms, line 14: Why `Multiline`? In the code below you are reading the file one line at a time...

src/ApplicationInsights.Kubernetes/ContainerIdProviders/MountInfoContainerIdProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.IO;
+using System.Text.RegularExpressions;
+using Microsoft.ApplicationInsights.Kubernetes.Debugging;
+
+namespace Microsoft.ApplicationInsights.Kubernetes.ContainerIdProviders;
+
+internal class MountInfoContainerIdProvider : IContainerIdProvider
+{
+    private const string InfoFilePath = "/proc/self/mountinfo";
+    private const string MatchPattern = @"\/docker\/containers\/(.*?)\/";
```
karolz-ms, line 13: Is this going to work for containerD and other container runtimes? Docker is not present by default with latest version of Kubernetes...

In general, how reliable is discovering container ID based on mount info? Is there a K8s spec/design doc that can be referenced here?

karolz-ms (commented):
I have some questions...

src/ApplicationInsights.Kubernetes/ContainerIdProviders/MountInfoContainerIdProvider.cs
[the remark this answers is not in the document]
xiaomi7732: Thanks for pointing out. Yes, slash is not required, but it works;
I put it there because it is the lower bar (works in js as well, for example).

1. I don't know there's any negative impact, or is there?
1. Do you think we I shall keep it very cleaning by remove the slashes? I believe remove them works.

xiaomi7732: @karolz-ms, I didn't know there's `containerD` and other runtimes (well, now I think about it, why isn't there? :-)). That is very good point. I need some time to learn it. Tracking by #275 .

I assume you are okay to unblock this change since:

1. It works with docker engine;
2. `ContainerIdProvider` is designed in a way of extensible - mountinfo provider is just one node on the change.

karolz-ms: Does it work with AKS? ☺️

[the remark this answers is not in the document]
karolz-ms: Thanks for the clarification! 

It is not a huge deal, but I can think of two negatives related to extra backslashes, both related to code readability:
1. They make the regexp harder to read than necessary.
2. They make me (and probably others) wonder why they are there, and whether they are significant.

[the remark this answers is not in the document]
xiaomi7732: Updated.

src/ApplicationInsights.Kubernetes/ContainerIdProviders/MountInfoContainerIdProvider.cs
xiaomi7732: Fixed.

src/ApplicationInsights.Kubernetes/ContainerIdProviders/MountInfoContainerIdProvider.cs
xiaomi7732: We synced offline. I am going to add another iteration to handle containerD as the runtime and we will leave other engines on request bases unless the usage changes.

xiaomi7732:
Hey @karolz-ms Thanks so much for the review. I really appreciated that you bring containerd into our attention. I've given it another iteration, mind take another review? Alone with the implementation, I also put up a draft for a wiki of the design we discussed, mind also give it a quick peak: <https://github.com/microsoft/ApplicationInsights-Kubernetes/wiki/Design---ContainerIdProviders-%5BDRAFT%5D> when you have time? Thanks.

src/ApplicationInsights.Kubernetes/ContainerIdProviders/EnvironmentVariableContainerIdProvider.cs
```diff
@@ -0,0 +1,24 @@
...
+using System;
+using Microsoft.ApplicationInsights.Kubernetes.Debugging;
+
+namespace Microsoft.ApplicationInsights.Kubernetes.ContainerIdProviders
+{
+    internal class EnvironmentVariableContainerIdProvider : IContainerIdProvider
+    {
+        private readonly ApplicationInsightsKubernetesDiagnosticSource _logger = ApplicationInsightsKubernetesDiagnosticSource.Instance;
+
+        private const string EnvironmentVariableName = "ContainerId";
```
karolz-ms, line 12: Consider making this a construction parameter, with `ContainerId` being a default

karolz-ms (approved):
Looks great1 Just one suggestion, but I am fine with leaving things as they are for now.

I also read through the design doc and it makes a lot of sense to me.

PR #283, opened 2022-06-17
Dev/saars/checking api access

xiaomi7732:
* Add a documentation about configuration service account permissions in RBAC enabled K8s clusters.
* Clean up permissions (verbs/apiGroups) needed.


docs/configure-rbac-permissions.md
```diff
@@ -0,0 +1,146 @@
...
+
+    ```yaml
+    ---
+    kind: ClusterRoleBinding
+    apiVersion: rbac.authorization.k8s.io/v1
+    metadata:
+    name: appinsights-k8s-property-reader-binding
+    subjects:
+    - kind: ServiceAccount
+    name: default
```
karolz-ms, line 38: name and namespace indentation do not seem right (should be part of the subject, I think)

docs/configure-rbac-permissions.md
```diff
@@ -0,0 +1,146 @@
...
+
+Kubectl provides an `auth --can-i` sub command for troubleshooting permissions. It supports impersonate the service account. We can leverage it for permission troubleshooting, for example:
+
+```shell
+kubectl auth can-i list pod --namespace default --as system:serviceaccount:default:default
+```
+
+There are several `default` there, let's break it down:
+
+* --namespace **default**: This is the target namespace, it is usually where your applications deploys to.
```
karolz-ms, line 67: I suggest using different namespace for the application (target namespace). Putting each application in a separate namespace is pretty much a standard practice in K8s world. It will make the instructions more realistic.

docs/configure-rbac-permissions.md
```diff
@@ -0,0 +1,146 @@
...
+The response for the command-line will simply be `yes` or `no`, for example:
+
+```shell
+$ kubectl auth can-i list pod --namespace default --as system:serviceaccount:default:default
+yes
+```
+
+## Use SubjectAccessReview
+
+Kubernetes also provides `SubjectAccessReview` to check permission for given user on target resource.
```
karolz-ms, line 79: Side note: did not know about `SubjectAccessReview`... I am not sure I like the pattern that this CRD uses, namely, "create an object and then wait for the controller to fill its status" as a way to make a request against the server. I think I would prefer an API aggregation approach instead. Oh well 🤷‍♂️

karolz-ms (changes requested):
Overall looks great, but I think we should use a separate namespace for the application, as this is the most common way of deploying apps on K8s... I am open to push back though ☺️

docs/configure-rbac-permissions.md
xiaomi7732: I think that is a good point! Let me work on it.

docs/configure-rbac-permissions.md
xiaomi7732: That was a #TIL for me too. One details though, when I try to query the objects after I created them, the query failed. I don't know if whether those objects really persistent there or are they just `temporary` to hold the status and return and immediately got disposed.

I like it b/c it give me a way to share a yaml for others to use.

docs/configure-rbac-permissions.md
xiaomi7732: I think it will make it more distinguishable too. Good point!

xiaomi7732:
@karolz-ms Thanks for the feedback. Instructions updated.

karolz-ms (approved):

Issue #276, opened 2022-06-06
Refactor: Make container id optional


xiaomi7732:
We have been using container id to locate pod name. It worked well except container id is not always within a container. In that case, allows a way to making the enhancement work without container id info is still preferrable.

xiaomi7732:
* Container id becomes optional for telemetries
    *  On Linux, since the container id is auto detected, when it is not available, a warning is issued;
        * There are 3 ways to get a container id included in the telemetry:
		1. The user explicit sets the environment variable of `ContainerId`.
            1. Auto detect.
            1. Fallback: There is only 1 container inside the pod, that will be used.
    *  On Windows, there will not be warnings since container id is optional. There are 2 ways for container id to be there:
		* 1 and 3 above.

Issue #172, opened 2019-01-18
Pod name requirement means extension doesn't "just work" on Windows


davidmatson:
Related to #171, unlike on Linux, at least one extra configuration step is required to get the extension working. This issue tracks making things "just work" on Windows like they do on Linux. Likely that requires driving changes upstream to make the pod name accessible from within the pod - where would we need to get assistance to make this work end-to-end?

xiaomi7732:
That is right.
In the Linux implementation, the pod name is parsed from the cgroup, which doesn't exist in the Windows world.
If anyone knows how to get the container id / pod info somehow from within a Windows container, I'll be glad to update the implementation to eliminate the step.

davidmatson:
Should we try asking the Kubernetes folks? https://github.com/kubernetes/kubernetes

davidmatson:
We tested and found that it is possible to get the host name on Windows from within the pod without using the downward API. Hostname.exe from within the pod works, which internally uses this Win32 API:
https://docs.microsoft.com/en-us/windows/desktop/api/winsock/nf-winsock-gethostname

Could this extension be updated to use this API on windows to have the pod name "just work" there too?

xiaomi7732:
@davidmatson Thanks for the info.
By reading the document for getHostName(), it returns the host name of the container, right? (like machine name or machine name in a domain)

I have difficulties to use that to correlate back to the container id or use that to figure out the pod name. **Could you please explain in detail how is that going to work**?

davidmatson:
Inside a container, this API returns the pod name. Try doing something like this:
kubectl exec <pod> hostname.exe
and compare it with
kubectl exec <pod> cmd /c echo %COMPUTERNAME%

xiaomi7732:
@davidmatson thanks for the quick turnaround. I'll take a look at it.

xiaomi7732:
@davidmatson Followed the lead, here's what I see:
```shell
saars@Saars-p51:~$ kubectl exec -it nanoserver-iis-5d999b87d7-mtz6w hostname.exe
c003088533e3
```
**c003088533e3** is the computer / host name, which agrees with the API spec.

**However**, it is neither the pod name nor can I correlate it back to the **pod name** (nanoserver-iis-5d999b87d7-mtz6w), what do I miss here?
```shell
saars@Saars-p51:~$ kubectl describe pod nanoserver-iis-5d999b87d7-mtz6w
[...]

PR #284, opened 2022-07-14
Dev/saars/optional container

xiaomi7732:
Address #276 and #172. Please refer to [design: Container id is optional while pod name required
](https://github.com/microsoft/ApplicationInsights-Kubernetes/wiki/Design:-Container-id-is-optional-while-pod-name-required) for designer ideas.


xiaomi7732:
See a blocker. Close the PR for now.

xiaomi7732:
Issue resolved. Resume the PR.

src/ApplicationInsights.Kubernetes/K8sQueryClient.cs
```diff
@@ -40,103 +43,102 @@ public K8sQueryClient(IKubeHttpClient kubeHttpClient)
...
+            _logger.LogTrace("Default Header: {0}", KubeHttpClient.DefaultRequestHeaders);
 
-            string myContainerId = KubeHttpClient.Settings.ContainerId;
-            K8sPod myPod = await GetMyPodAsync().ConfigureAwait(false);
-            return myPod.GetContainerStatus(myContainerId);
+            using HttpResponseMessage response = await KubeHttpClient.SendAsync(request, cancellationToken).ConfigureAwait(false);
+            if (response.IsSuccessStatusCode)
+            {
+                _logger.LogTrace("Query succeeded.");
+                string resultString = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
```
karolz-ms, line 113:
```suggestion
                string resultString = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
```

src/ApplicationInsights.Kubernetes/PodInfoProviders/PodInfoManager.cs
```diff
@@ -0,0 +1,100 @@
...
+        _k8SQueryClientFactory = k8SQueryClientFactory ?? throw new ArgumentNullException(nameof(k8SQueryClientFactory));
+        _settingsProvider = settingsProvider ?? throw new System.ArgumentNullException(nameof(settingsProvider));
+        _podNameProviders = podNameProviders ?? throw new System.ArgumentNullException(nameof(podNameProviders));
+    }
+
+    /// <inheritdoc />
+    public async Task<K8sPod?> GetMyPodAsync(CancellationToken cancellationToken)
+    {
+        string podName = string.Empty;
+        foreach (IPodNameProvider podNameProvider in _podNameProviders)
```
karolz-ms, line 35: I wonder if it would be possible (and desirable) to use `Enumerable.Any()` instead of explicit loop here?

karolz-ms (approved):
Wow, big change. Very nice! 🎉 Just a few minor questions

xiaomi7732:
> Wow, big change. Very nice! 🎉 Just a few minor questions

Yup, thanks. TBH, I was halfway through and found myself in a rabbit hole :-)

src/ApplicationInsights.Kubernetes/PodInfoProviders/PodInfoManager.cs
xiaomi7732: > I wonder if it would be possible (and desirable) to use `Enumerable.Any()` instead of explicit loop here?

Yes, it is possible but I did consider it and chose not to do it. Here are my considerations:

* there is a precedence when both variables are provided
    *  I actually don't see a reason for the user to overwrite the podName by variable unless the $HOSTNAME is somehow messed up; And should that happen, user choice win.
* I guess the perf gain is very very very subtle :-)

Let me know if I miss anything.

src/ApplicationInsights.Kubernetes/PodInfoProviders/PodInfoManager.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        // Find pod by container id:
+        string myContainerId = _settingsProvider.ContainerId;
+        _logger.LogDebug($"Looking for pod name by container id: {myContainerId}");
+
+        if (!string.IsNullOrEmpty(myContainerId))
+        {
+            K8sPod? targetPod = null;
+            using (IK8sQueryClient k8sQueryClient = _k8SQueryClientFactory.Create())
```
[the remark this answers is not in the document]
xiaomi7732: That has been considered :-)
I think the gain is what you mentioned - clearer code; and the pro for it here we dispose the k8sQueryClient slightly earlier than the scope of the if statement.
To me seems like a tomato tomato. :-D Let me know if I miss anything.

src/ApplicationInsights.Kubernetes/K8sQueryClient.cs
xiaomi7732: Yup. Unfortunately, the overload that takes in the cancellation token is not there in Netstandard 2.0 :-)

https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpcontent.readasstringasync?view=netstandard-2.0

Issue #292, opened 2022-08-12
Container id found by cgroup requires normalization


xiaomi7732:
An issue related to #290, reported by [SantoshG-Derivco](https://github.com/SantoshG-Derivco).

TL;DR
* In cGroupV1, **container id** info: 6:cpuset:/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod3e221707_600c_4eb1_8fe7_a9a7163a1278.slice/cri-containerd-5146b2bcd77ab4f2624bc1fbd98cf9751741344a80b043dbd77a4e847bff4f06.scope
* The **matched** container id: cri-containerd-**5146b2bcd77ab4f2624bc1fbd98cf9751741344a80b043dbd77a4e847bff4f06**.scope
* The **expected** container id: 5146b2bcd77ab4f2624bc1fbd98cf9751741344a80b043dbd77a4e847bff4f06

xiaomi7732:
- [ ] Fix in 2.x
- [ ] Fix in 3.x

xiaomi7732:
This issue should have been addressed in [2.0.5-beta1](https://www.nuget.org/packages/Microsoft.ApplicationInsights.Kubernetes/2.0.5-beta1). Please check it out.

I'll keep this issue open until the same functionality has been forward port to 3.0.

SantoshG-Derivco:
@xiaomi7732 I've tested [2.0.5-beta1](https://www.nuget.org/packages/Microsoft.ApplicationInsights.Kubernetes/2.0.5-beta1) and all looks good thanks 😄

xiaomi7732:
@SantoshG-Derivco Thanks for the confirmation. It is important for us to know that it works for you :-).

xiaomi7732:
Stable package released:
https://www.nuget.org/packages/Microsoft.ApplicationInsights.Kubernetes/2.0.5

PR #293, opened 2022-08-12
Dev/saars/normalize container

xiaomi7732:
Address #292 

To address the potential mismatch that could happen for all container id providers, I introduced a new interface - `IContainerIdNormalizer` to normalize all formats of container ids to the GUID/UUID only. In the example of the issue, it would be `5146b2bcd77ab4f2624bc1fbd98cf9751741344a80b043dbd77a4e847bff4f06` to match the expectation.

The normalization always happens after the container id is fetched.


karolz-ms (approved):
Looks good ☺️

PR #294, opened 2022-08-12
Refresh worker example

xiaomi7732:
* Recommend deploying to custom namespace
* Link to RBAC configuration
* Some small changes


karolz-ms (approved):
Looks good! Consider using a fixed version tag for images, `1.0` or something, the tend to work better in practice, e.g. see https://vsupalov.com/docker-latest-tag/

PR #300, opened 2022-08-26
Using background service to fetch k8s cluster info

xiaomi7732:
Addresses #301, #302 

Leverage background service for getting cluster-info.
  * Simplified the telemetry initializer
  * Enables refresh cluster-info by configured intervals.

Reimplement timeout by using a cancellation token than a stub task.

Piggyback some small refactors.

### DIFF
--- a/src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
+++ b/src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
@@ -38,15 +38,15 @@ namespace Microsoft.ApplicationInsights.Kubernetes
         /// Async factory method to build the instance of a K8sEnvironment.
         /// </summary>
         /// <returns></returns>
-        public async Task<IK8sEnvironment?> CreateAsync(DateTime timeoutAt, CancellationToken cancellationToken)
+        public async Task<IK8sEnvironment?> CreateAsync(CancellationToken cancellationToken)
         {
             try
             {
                 // When my pod become available and it's status become ready, we recognize the container is ready.
 
                 // TODO: See if there's better way to fetch pod
-                V1Pod myPod = await SpinWaitUntilGetPodAsync(timeoutAt, cancellationToken).ConfigureAwait(false);
-                V1ContainerStatus? containerStatus = await SpinWaitContainerReadyAsync(timeoutAt, cancellationToken).ConfigureAwait(false);
+                V1Pod myPod = await SpinWaitUntilGetPodAsync(cancellationToken).ConfigureAwait(false);
+                V1ContainerStatus? containerStatus = await SpinWaitContainerReadyAsync(cancellationToken).ConfigureAwait(false);
 
                 // Fetch replica set info
                 IEnumerable<V1ReplicaSet> allReplicaSet = await _k8sClient.GetReplicaSetsAsync(cancellationToken).ConfigureAwait(false);
@@ -77,13 +77,15 @@ namespace Microsoft.ApplicationInsights.Kubernetes
 #pragma warning restore CA1031 // Do not catch general exception types
         }
 
-        private async Task<V1Pod> SpinWaitUntilGetPodAsync(DateTime timeoutAt, CancellationToken cancellationToken)
+        private async Task<V1Pod> SpinWaitUntilGetPodAsync(CancellationToken cancellationToken)
         {
             Stopwatch stopwatch = new Stopwatch();
             stopwatch.Start();
             V1Pod? myPod = null;
             do
             {
+                cancellationToken.ThrowIfCancellationRequested();
+
                 try
                 {
                     myPod = await _podInfoManager.GetMyPodAsync(cancellationToken).ConfigureAwait(false);
@@ -107,28 +109,26 @@ namespace Microsoft.ApplicationInsights.Kubernetes
                 // Try invoke a probe on readiness every 500ms until the container is ready
                 // Or it will timeout per the timeout settings.
                 await Task.Delay(500).ConfigureAwait(false);
-            } while (DateTime.Now < timeoutAt);
-
-            _logger.LogDebug($"{nameof(SpinWaitUntilGetPodAsync)} timed out.");
-            throw new InvalidOperationException("Can't find pod information in given time.");
+            } while (true);
         }
 
         /// <summary>
         /// Waits until the container is ready.
         /// Refer document @ https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase for the Pod's lifecycle.
         /// </summary>
-        /// <param name="timeoutAt">A point in time when the wait on container startup is abandoned.</param>
         /// <param name="cancellationToken">The cancellation token.</param>
         /// <returns>
         /// Returns the pod with refreshed info on it.
         /// </returns>
-        private async Task<V1ContainerStatus?> SpinWaitContainerReadyAsync(DateTime timeoutAt, CancellationToken cancellationToken)
+        private async Task<V1ContainerStatus?> SpinWaitContainerReadyAsync(CancellationToken cancellationToken)
         {
             Stopwatch stopwatch = new Stopwatch();
             stopwatch.Start();
 
             do
             {
+                cancellationToken.ThrowIfCancellationRequested();
+
                 try
                 {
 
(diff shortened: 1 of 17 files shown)


src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
```diff
@@ -149,8 +149,7 @@ private async Task<V1Pod> SpinWaitUntilGetPodAsync(DateTime timeoutAt, Cancellat
                 // Try invoke a probe on readiness every 500ms until the container is ready
                 // Or it will timeout per the timeout settings.
                 await Task.Delay(500).ConfigureAwait(false);
```
xiaomi7732, line 151: Use Delay with cancellation token

src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
```diff
@@ -107,28 +109,26 @@ private async Task<V1Pod> SpinWaitUntilGetPodAsync(DateTime timeoutAt, Cancellat
...
```
xiaomi7732, line 111: Use Delay with cancellation token

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        while (true)
+        {
+            try
+            {
+                _logger.LogDebug($"Starting {nameof(UpdateK8sEnvironmentAsync)}");
+                await UpdateK8sEnvironmentAsync(stoppingToken).ConfigureAwait(false);
+                TimeSpan interval = _options.ClusterInfoRefreshInterval;
+                _logger.LogDebug($"Finished {nameof(UpdateK8sEnvironmentAsync)}, delay for {interval}");
+                await Task.Delay(interval);
```
xiaomi7732, line 41: Delay with cancellation token

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
```diff
@@ -0,0 +1,90 @@
...
+            }
+#pragma warning restore CA1031 // Do not catch general exception types
+        }
+    }
+
+    private async Task UpdateK8sEnvironmentAsync(CancellationToken cancellationToken)
+    {
+        await using (AsyncServiceScope scope = _serviceProvider.CreateAsyncScope())
+        {
+            // Get scoped service ready
```
xiaomi7732, line 58: services

src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
```diff
@@ -107,28 +109,26 @@ private async Task<V1Pod> SpinWaitUntilGetPodAsync(DateTime timeoutAt, Cancellat
...
         /// </summary>
-        /// <param name="timeoutAt">A point in time when the wait on container startup is abandoned.</param>
         /// <param name="cancellationToken">The cancellation token.</param>
         /// <returns>
         /// Returns the pod with refreshed info on it.
         /// </returns>
-        private async Task<V1ContainerStatus?> SpinWaitContainerReadyAsync(DateTime timeoutAt, CancellationToken cancellationToken)
+        private async Task<V1ContainerStatus?> SpinWaitContainerReadyAsync(CancellationToken cancellationToken)
         {
             Stopwatch stopwatch = new Stopwatch();
```
karolz-ms, line 125: This Stopwatch does not seep to be used anywhere?

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
```diff
@@ -0,0 +1,90 @@
...
+        _options = options?.Value ?? throw new ArgumentNullException(nameof(options));
+    }
+
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        while (true)
+        {
+            try
+            {
+                _logger.LogDebug($"Starting {nameof(UpdateK8sEnvironmentAsync)}");
```
karolz-ms, line 37: If `stoppingToken` is never checked, how will this service shut down?

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
```diff
@@ -0,0 +1,90 @@
...
+                    IK8sEnvironment? environment = await factory.CreateAsync(cancellationToken: linkedTokeSource.Token).ConfigureAwait(false);
+                    _k8SEnvironmentHolder.K8sEnvironment = environment;
+                }
+            }
+            catch (OperationCanceledException)
+            {
+                if (timeoutToken.IsCancellationRequested)
+                {
+                    _logger.LogWarning("Querying Kubernetes cluster info timed out.");
+                    _k8SEnvironmentHolder.K8sEnvironment = null;
```
karolz-ms, line 80: This is probably fine because `IsQueryTimeout` will only be set once, and `K8sEnvironment` will never be anything other than `null` if `IsQueryTimeout == true`.

In general though, it gets pretty tricky when one thread is setting a few related properties, and other threads need to read all values in a consistent way. The only way to ensure correctness in such (general) case is to use locks or other concurrency mechanisms.

karolz-ms (changes requested):
Overall looks pretty good, but I have some questions

src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
xiaomi7732: Probably from legacy code - I'll double-check.

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
xiaomi7732: Of course :-) I'll update the code.

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
xiaomi7732: Got rid of IsQueryTimeout already. I want something very very simple. It is too easy to make it more complex than needed.

xiaomi7732: I was expecting that question :-) I want to discuss it. That is actually something I considered and here are my thoughts:

In general, I want to avoid complexity, so, I designed it a bit:

1. The K8sEnvHolder is a singleton, and it holds a ref to K8sEnv object - properties on K8sEnv doesn't mutate;
2. When we gather new info for K8sEnv, it builds a new K8sEnv object, and when it is ready, update the ref - [code](https://github.com/xiaomi7732/ApplicationInsights-Kubernetes/blob/72043d4cf9ea3ec58a4ccdea20ea43aa62c55cc7/src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs#L72);
3. On the consuming side (initializer), the first thing we do is to reference the current K8sEnv object and always use it - [code](https://github.com/xiaomi7732/ApplicationInsights-Kubernetes/blob/72043d4cf9ea3ec58a4ccdea20ea43aa62c55cc7/src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs#L37).

In general, I designed it that way to avoid unnecessary semaphores/locks. Do you think that would work? Or do you see anything that is missing?

src/ApplicationInsights.Kubernetes/K8sEnvironmentFactory.cs
xiaomi7732: Removed.

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
xiaomi7732: Thanks for pointing out.

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
karolz-ms: Yes, I think it will work. If you want to be 100% safe with lock-free code like this, make the reference to K8sEnvironment within the K8sEnvironmentHolder a volatile field.

karolz-ms (approved):

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
karolz-ms: (unresolving to make sure you saw the previous comment, feel free to re-resolve)

xiaomi7732: volatile it is.

Issue #303, opened 2022-08-30
Support SDK without ClusterRole permissions


Basvdlouw:
Currently the SDK only works within an RBAC environment when a ClusterRole is defined. Within our setup we cannot define ClusterRoles. 
We would like to use regular Roles/RoleBindings. The "nodes" resource requires ClusterRole permissions, but the rest of the resources that are logged do not. We should still be able to log this information to app insights instead of getting an unauthorized exception on initialization.

xiaomi7732:
@Basvdlouw Thanks for the feedback. It is very specific and that is helpful. Let me do some investigation around it.

xiaomi7732:
Hey @Basvdlouw, it turned out to be a minor tweak and the change is released in: https://www.nuget.org/packages/Microsoft.ApplicationInsights.Kubernetes/2.0.6-beta1
Mind giving it a try to see if that works for you?

Here's an example we used to set up the RoleBinding: https://github.com/microsoft/ApplicationInsights-Kubernetes/blob/develop/docs/sa-role-none-cluster.yaml

Again, thanks for your contribution!

Basvdlouw:
Hi @xiaomi7732, I've tested it and it works as expected.
Thanks for the quick fix

PR #305, opened 2022-09-01
Support telemetry enhancement without cluster role

xiaomi7732:
Addresses #303 for 2.x.

### DIFF
--- a/docs/configure-rbac-permissions.md
+++ b/docs/configure-rbac-permissions.md
@@ -1,19 +1,19 @@
 # Configure RBAC permissions
 
-`Microsoft.ApplicationInsights.Kubernetes` uses the service account to query kubernetes information to enhance telemetries. It is important to have proper permissions configured for kubernetes related information like node, pod and so on to be fetched correctly.
+`Microsoft.ApplicationInsights.Kubernetes` uses the service account to query Kubernetes information to enhance telemetries. It is important to have proper permissions configured for Kubernetes-related information like node, pod, and so on to be fetched correctly.
 
-In this post, we will start by describe a method to correctly configure the permissions for an RBAC enabled cluster. And then share a guidance for troubleshooting.
+In this post, we will start by describing a method to correctly configure the permissions for an RBAC-enabled cluster. And then share troubleshooting guidance.
 
 ## Assumptions
 
 In this demo, we will have the following assumptions. Please change the related values accordingly:
 
-* The application will be deployed to namespace of `ai-k8s-demo`.
+* The application will be deployed to the namespace of `ai-k8s-demo`.
 * The application will leverage the `default` service account.
 
 ## Setup the permissions for the service account
 
-Depends on various considerations, there could be different strategies to setup the permissions for your service account. Here we list 2 common possibilities, as examples.
+Depending on various considerations, there could be different strategies to set up the permissions for your service account. Here we list 2 common possibilities, as examples.
 
 * If you want to get the node information along with other info like pod, deployment, and so on, a ClusterRole and a ClusterRoleBinding are required, and here's how to do it:
 
@@ -55,7 +55,7 @@ Depends on various considerations, there could be different strategies to setup 
             apiGroup: rbac.authorization.k8s.io
           ```
 
-      That is to grant the role of `appinsights-k8s-property-reader` to the default service account in namespace of `ai-k8s-demo`.
+      That is to grant the role of `appinsights-k8s-property-reader` to the default service account in the namespace of `ai-k8s-demo`.
 
   * If you don't want to create a Cluster Role, it is also possible to use Role and RoleBinding starting 2.0.6+. Follow the example in [sa-role-none-cluster.yaml](./sa-role-none-cluster.yaml). In that case, you will not have node info on the telemetries.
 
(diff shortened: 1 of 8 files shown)


docs/configure-rbac-permissions.md
```diff
@@ -1,70 +1,76 @@
 # Configure RBAC permissions
 
-`Microsoft.ApplicationInsights.Kubernetes` uses the service account to query kubernetes information to enhance telemetries. It is important to have proper permissions configured for kubernetes related information like node, pod and so on to be fetched correctly.
+`Microsoft.ApplicationInsights.Kubernetes` uses the service account to query Kubernetes information to enhance telemetries. It is important to have proper permissions configured for Kubernetes-related information like node, pod, and so on to be fetched correctly.
```
karolz-ms, line 3:
```suggestion
`Microsoft.ApplicationInsights.Kubernetes` uses the service account to query Kubernetes information to enhance telemetries. It is important to have proper permissions configured for Kubernetes-related resources like Node, Pod, and so on to be fetched correctly.
```
(in K8s API lingo Node, Pod, Deployment etc are called "resources" and spelled starting with a capital letter)

docs/configure-rbac-permissions.md
```diff
@@ -1,70 +1,76 @@
...
 
-In this post, we will start by describe a method to correctly configure the permissions for an RBAC enabled cluster. And then share a guidance for troubleshooting.
+In this post, we will start by describing a method to correctly configure the permissions for an RBAC-enabled cluster. And then share troubleshooting guidance.
 
 ## Assumptions
 
 In this demo, we will have the following assumptions. Please change the related values accordingly:
 
-* The application will be deployed to namespace of `ai-k8s-demo`.
+* The application will be deployed to the namespace of `ai-k8s-demo`.
```
karolz-ms, line 11:
```suggestion
* The application will be deployed to namespace `ai-k8s-demo`.
```

docs/configure-rbac-permissions.md
```diff
@@ -1,70 +1,76 @@
...
-          apiGroup: rbac.authorization.k8s.io
-        ```
-
-    That is to grant the role of `appinsights-k8s-property-reader` to the default service account in namespace of `ai-k8s-demo`.
-    
+## Setup the permissions for the service account
+
+Depending on various considerations, there could be different strategies to set up the permissions for your service account. Here we list 2 common possibilities, as examples.
+
+* If you want to get the node information along with other info like pod, deployment, and so on, a ClusterRole and a ClusterRoleBinding are required, and here's how to do it:
```
karolz-ms, line 18:
```suggestion
* If you want to get the Node information along with other resource info like Pod, Deployment, and so on, a ClusterRole and a ClusterRoleBinding are required, and here's how to do it:
```

docs/configure-rbac-permissions.md
```diff
@@ -1,70 +1,76 @@
...
+            namespace: ai-k8s-demo
+          roleRef:
+            kind: ClusterRole
+            name: appinsights-k8s-property-reader
+            apiGroup: rbac.authorization.k8s.io
+          ```
+
+      That is to grant the role of `appinsights-k8s-property-reader` to the default service account in the namespace of `ai-k8s-demo`.
+
+  * If you don't want to create a Cluster Role, it is also possible to use Role and RoleBinding starting 2.0.6+. Follow the example in [sa-role-none-cluster.yaml](./sa-role-none-cluster.yaml). In that case, you will not have node info on the telemetries.
```
karolz-ms, line 60: What is the `2.0.6` version referring to? Assume that is the version of the ApplicationInsights Kubernetes NuGet package? It would be worth to clarify.

docs/configure-rbac-permissions.md
```diff
@@ -1,70 +1,76 @@
...
+
 * Now you can deploy it:
 
     ```shell
     kubectl create -f `sa-role.yaml`
     ```
     See [sa-role.yaml](sa-role.yaml) for a full example.
 
-> :warning: Check back for various permissions needed. Depends on the implementations, it may change in the over time.
+> :warning: Check back for various permissions needed. Depending on the implementations, it may change over time.
```
karolz-ms, line 69: Implementation of what? Can you clarify?

src/ApplicationInsights.Kubernetes/K8sQueryClient.cs
```diff
@@ -91,12 +91,27 @@ public Task<IEnumerable<K8sDeployment>> GetDeploymentsAsync(CancellationToken ca
...
+            {
+                // Prefer ignoring the unauthorized access exception
+                if (ignoreForbiddenException)
+                {
+                    _logger.LogDebug(ex.Message);
+                    _logger.LogTrace(ex.ToString());
+                    return Enumerable.Empty<K8sNode>();
+                }
+                // Otherwise
+                throw;
```
karolz-ms, line 113: Nit: you can just say `else throw;`. No comment necessary ☺️

karolz-ms (approved):
Code looks great. I have a few suggestions related to the docs, but nothing super important

docs/configure-rbac-permissions.md
xiaomi7732: Cool

src/ApplicationInsights.Kubernetes/K8sQueryClient.cs
xiaomi7732: I am sure you are okay with that, I just want to call out in case the return was oversighted. Thanks for the suggestion anyway.

PR #309, opened 2022-09-08
Clean up lifetime of registered services

xiaomi7732:
Address #308


karolz-ms (approved):
Looks great.

(Saar explained to me in person that the "scoped" services are scoped to the K8s information refresh loop and are not directly used by the K8s telemetry initializer, so the are not instantiated per-request)

PR #311, opened 2022-09-09
Refactor: Hide public interfaces, clean up namespace

xiaomi7732:
Addresses #310 

* Hide 2 pieces of unintentionally exposed public interfaces;
* Simplify the namespaces;

No functional change is expected.


src/ApplicationInsights.Kubernetes/Extensions/KubernetesServiceCollectionBuilder.cs
```diff
@@ -2,142 +2,140 @@
...
+            _logger.LogError("Unsupported OS.");
         }
+
+        serviceCollection.TryAddScoped<IContainerIdNormalizer, ContainerIdNormalizer>();
+
+        // Notes: pay attention to the order. Injecting uses the order of registering in this case.
+        // For backward compatibility, $APPINSIGHTS_KUBERNETES_POD_NAME has been agreed upon to allow customize pod name with downward API.
+        serviceCollection.TryAddEnumerable(ServiceDescriptor.Scoped<IPodNameProvider, UserSetPodNameProvider>());
+        // $Hostname will be overwritten by Kubernetes to reveal pod name: https://kubernetes.io/docs/concepts/containers/container-environment/#container-information.
+        serviceCollection.TryAddEnumerable(ServiceDescriptor.Scoped<IPodNameProvider, HostNamePodNameProvider>());
```
JacobBovee, line 129: Not relevant to this PR specifically but I'm curious about why you're adding a provider class for each possible environment variable. It seems like a lot of added ceremony unless those managers are expected to get a lot more complex. 

I wonder if you could do the same thing with a readonly queue in the `EnvironmentVariablePodNameProviderBase.cs` class and remove the individual implementations.

JacobBovee (approved):
Looks good to me. Mostly nits so feel free to ignore.

src/ApplicationInsights.Kubernetes/Containers/ContainerIdProviders/IContainerIdProvider.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace Microsoft.ApplicationInsights.Kubernetes.ContainerIdProviders
+﻿namespace Microsoft.ApplicationInsights.Kubernetes.Containers
 {
```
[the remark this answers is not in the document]
xiaomi7732: Yup. I missed the spot

src/ApplicationInsights.Kubernetes/Extensions/KubernetesServiceCollectionBuilder.cs
```diff
@@ -2,142 +2,140 @@
...
+    /// Register Kubernetes related service into the service collection.
+    /// </summary>
+    /// <param name="serviceCollection">The service collector to inject the services into.</param>
+    /// <returns>Returns the service collector with services injected.</returns>
+    public IServiceCollection RegisterServices(IServiceCollection serviceCollection)
+    {
+        if (_clusterCheck.IsInCluster)
         {
-            if (_clusterCheck.IsInCluster)
+            if (serviceCollection == null)
```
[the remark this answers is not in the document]
xiaomi7732: Sure

src/ApplicationInsights.Kubernetes/Extensions/KubernetesServiceCollectionBuilder.cs
xiaomi7732: That is a very good question.

My rationale behind it is the OC principal. By code history, UserSetPodName was implemented way earlier than the HostName. When the time hostname was implemented later, I had two choices:
1. Change the existing implementation of the user set environment variable class and update code there;
2. Extend the implementation by adding another class.

In this case, I actually am okay with either choice. Nothing complex at all and I don't expect any bugs.
For more complex scenarios, though, I would prefer a new implementation to change the existing code.

Does that answer the question? Please let me know and we can have a further discuss on it.

karolz-ms (approved):

src/ApplicationInsights.Kubernetes/Extensions/KubernetesServiceCollectionBuilder.cs
JacobBovee: Totally makes sense. Probably not worth spending time on unless it comes up more. Thanks for the answer.

Issue #325, opened 2022-10-20
Application Insights for Kubernetes 3.0: telemetry initializer is used but the content has not ready yet in Console Applications


qbdev-kr:
### Goal

I want to be able to see Enriched Kubernetes Info coming into App Insights

### Issue Summary
I consistently am seeing the trace, 

**Application Insights for Kubernetes telemetry initializer is used but the content has not ready yet.**
![image](https://user-images.githubusercontent.com/114493252/197005241-097f3b03-8fa5-463d-8897-311067bfcbc7.png)

### What I have done
I have used https://github.com/microsoft/ApplicationInsights-Kubernetes/issues/187 to help me out by adding delay etc. But that has not helped me and i am also deploying to AKS.

### AKS
AKS is used to host a docker image from this code. The console app is calling an await method to run some ETL process and the total job completes in 2 minutes.

### Usage
The code snipper example below is identical to the [BasicConsoleAppILogger](https://github.com/microsoft/ApplicationInsights-Kubernetes/tree/develop/examples/BasicConsoleAppILogger)Application in .NET 6 console application with all needed packages installed.

```
const string ApplicationInsightsInstrumentationKey = "InstrumentationKey=<redacted keyValue from App Insights>";



IServiceCollection services = new ServiceCollection();
services.AddSingleton<Microsoft.Extensions.Configuration.IConfiguration>(p => new ConfigurationBuilder().Build());
var observer = new ApplicationInsightsKubernetesDiagnosticObserver(DiagnosticLogLevel.Trace);
ApplicationInsightsKubernetesDiagnosticSource.Instance.Observable.SubscribeWithAdapter(observer);

services.AddApplicationInsightsKubernetesEnricher(diagnosticLogLevel: LogLevel.Information);

var channel = new InMemoryChannel();
services.AddOptions<TelemetryConfiguration>().Configure<IServiceProvider>((config, p) =>
{
    // Appending registered telemetry initializers.
    foreach (ITelemetryInitializer registeredInitializer in p.GetServices<ITelemetryInitializer>())
    {
        config.TelemetryInitializers.Add(registeredInitializer);
    }

    // Setup up telemetry channels.
    config.TelemetryChannel = channel;
});

services.AddLogging(builder =>
{
    builder.AddFilter<ApplicationInsightsLoggerProvider>("", LogLevel.Trace);
    
    // Below is a different way to set Key to use the non-obselete extension method for AddApplicationInsights.
    
[...]

PR #326, opened 2022-10-20
Support manual bootstrap in console application

xiaomi7732:
Addresses #325.

### Problem
`Microsoft.ApplicationInsights.Kubernetes.3.x` leverages the hosted background service to fetch Kubernetes cluster properties. When there is no hoster, a plain console application, for example, there's no code to start the fetch of the cluster-info.

### Solution
Provides another extension method that can bootstrap the fetch of cluster-info.

```csharp
// Bootstrap k8s cluster-info. Only run this once.
serviceProvider.StartApplicationInsightsKubernetesEnricher();
```

### Alterntaives considered
- Expose a loose service for the user to get out of the IoC container - not as convenient; 1 additional public interface;
- Fully support non-IoC scenario - big change with little gain.

### Trade-offs

* For ConsoleApplication author, an additional line of code become required - StartApplicationInsightsKubernetesEnricher();
* Under some circumstances, it might bring confusion for non-console users.

### Tests
- [x] Unit tests pass.
- [x] Console app reporting cluster-info.
- [x] F5 project deployed to non-k8s - no crash, proper logging.
- [x] F5 project deployed to k8s - cluster info fetched correctly.


src/ApplicationInsights.Kubernetes/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -40,6 +42,20 @@ public static partial class ApplicationInsightsExtensions
             return services;
         }
 
+        /// <summary>
+        /// Bootstraps Application Insights Kubernetes enricher. This is intended to be used in console application, where hosted service are not invoked supported by default.
```
karolz-ms, line 46:
```suggestion
        /// Bootstraps Application Insights Kubernetes enricher. This is intended to be used in console application that does not use hosted services.
```

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
```diff
@@ -30,6 +30,26 @@ internal class K8sInfoBackgroundService : BackgroundService
...
+    void IK8sInfoBootstrap.Run()
+    {
+        try
+        {
+            // Fire and forget on purpose to avoid blocking the client code's thread.
+            _ = Task.Run(async () =>
+            {
+                using (CancellationTokenSource cancellationTokenSource = new CancellationTokenSource())
+                {
+                    await ExecuteAsync(cancellationTokenSource.Token).ConfigureAwait(false);
```
karolz-ms, line 43: `CancellationToken.None`?

Or consider passing in a `CancellationToken` and allowing to cancel the bootstrapping process

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
```diff
@@ -30,6 +30,26 @@ internal class K8sInfoBackgroundService : BackgroundService
         _options = options?.Value ?? throw new ArgumentNullException(nameof(options));
     }
 
+    // Allows non-hosted service to bootstrap cluster info fetch.
+    void IK8sInfoBootstrap.Run()
```
karolz-ms, line 34: Can we make this idempotent so it is less error-prone? I.e.
- OK for user code to call this multiple times (the cluster info is guaranteed to be initialized only once)
- OK to call it even if service host is present

This should be relatively easy to achieve using `Lazy<T>` or `LazyInitializer`.

karolz-ms (commented):
This is moving in the right direction but I wonder if we can make it more robust

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
xiaomi7732: Let me see if None is an option since you pointed it out.
I didn't pay too much effort there because I think that are implementation details and we can always improve later.

src/ApplicationInsights.Kubernetes/Extensions/ApplicationInsightsExtensions.cs
xiaomi7732: :-) Thanks.

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
xiaomi7732: Exactly what I was thinking too. The only reason I didn't put it out was that I am worried that the current implementation might not be the most optimal one, and I want to first release a beta package for the customer to try and see if they like the signature or not before working on improving the details.
I think I can have another PR to improve the implementations, what do you think?

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
karolz-ms: The ability to cancel bootstrapping (if introduced) would be a contract change, not just implementation detail. Also, ability to shut down gracefully and timely is generally very desirable.

xiaomi7732: Used `default` to supply an empty cancellation token for this scenario.

src/ApplicationInsights.Kubernetes/K8sInfoBackgroundService.cs
xiaomi7732: It turned out to be a bit more complex than that. For Lazy<Task>, it takes more than just that to have a cancellation token. An alternative that I see is to inherit from `Lazy<T>` to form a `Lazy<Task>`, having cancellation token support there, it is more effort than this bug fix. I will file another issue to track it.

Issue #327, opened 2022-10-21
[Investigation] Make calling into ExecuteAsync idempotent


xiaomi7732:
Tag @karolz-ms. He initiated the idea.

In the current implementation, it is undesire to call `StartApplicationInsightsKubernetesEnricher()` when hosted services exist, because it leads to duplicated call into ExecuteAsync - 1 from the background service, 1 from the explicit call of the method above.

Look into how to make `ExecuteAsync()` idempotent or find out a way to allow `ExecuteAsync()` to be called multiple times - that will reduce the chance for client code to call `StartApplicationInsightsKubernetesEnricher()` by accident, and become more refactor friendly.

xiaomi7732:
Notes:
1. Lazy<Task> for ExecuteAsync() - that the task never creates a second time: The issue is cancellation token handling. Lazy<T> doesn't take in a cancellation token; To implement it, it may be a `Lazy<Func<CancellationToken, Task>>`, and then, it starts to look like an AsyncLazy library - that feels a bit heavy for this case; It is also an intention not to own tricky implementations in this library.

2. Use an AsyncLazy library - don't need to worry about owning the complex logic. I went over some open-source libraries, and most of them are more than just AsyncLazy, and not all libraries are always dealing with all the issues. It is doable, it is probably not worth introducing the dependency for this scenario.

3. Going back to the origin of the issue - do not run the code 2nd time, and a stupidly simple solution would be: Have a ref counter, start at 0, and drop the execution when it is increased to 1. Here's a pseudo:

```csharp
int _refCount = 0;

async Task ExecuteAsync(CancenllationToken cancellationToken)
{
    if ( refCount >= 1 ) { return; } // Already executed;
    Interlocked.Increment(ref refCount);
    if ( refCount != 1 ) { return; } // Only thread got refCount == 1 survives.
    // Start to run the logic. No cancellation token passing around.
}
```

4. Introduce a SemaphoreSlim(1,1), one thread grabs the only available resource and never releases it, anything coming in afterward got rejected. - seems a bit to use SemaphoreSlim by force.

I am currently leaning toward 3. I think it is simple enough and easy to understand the semantics, and it doesn't have additional dependencies and is lightweight.

xiaomi7732:
Another approach suggested by @karol-ms: just take a lock, considering this isn't any hot path

```csharp
static readonly _lock = new object();
bool _executed;

[...]

PR #328, opened 2022-10-25
Make bootstrap idempotent

xiaomi7732:
Addresses #327. Piggybacked some simple refactor.

1. Flipped the dependency.
    * It was `IK8sInfoBootstrap` depends on `K8sInfoBackgroundService`
    * Changed to `K8sInfoBackgroundService` depends on `K8sInfoBootstrap`
    So that logic has been pulled out of `K8sInfoBackgroundService`. The responsibility for `K8sInfoBackgroundService` is clearly to handle background service startup.
    And the bootstrap logic is shared by both the background service and called by the client directly.

1. The fetch of the cluster properties is pulled out of the bootstrap implementations to divide the responsibilities:
    * `K8sInfoBootstrap` is responsible for handling bootstrap logic - initial delay, time out, query intervals, etc.
    * `K8sEnvironmentFetcher` is responsible for creating a new K8sEnv object and setting it to K8sEnvHolder.
    It is more aligned with the single responsibility principle, isolating those responsibilities also made it possible to mock half of the responsibilities and test the other half.

1. Added unit tests to cover DI container service resolution.

Tests
- [x] F5 project E2E - deployed to k8s, got the k8s properties.
- [x] Console application deployed to k8s, got the k8s properties.


src/ApplicationInsights.Kubernetes/K8sInfoBootstrap.cs
```diff
@@ -0,0 +1,87 @@
...
+                _logger.LogError("Unexpected error happened. Telemetry enhancement with Kubernetes info won't happen. Message: {0}", ex.Message);
+                _logger.LogTrace("Unexpected error happened. Telemetry enhancement with Kubernetes info won't happen. Details: {0}", ex.ToString());
+            }
+#pragma warning restore CA1031 // Do not catch general exception types
+        }
+    }
+
+
+    // Allows non-hosted service to bootstrap cluster property fetch.
+    public void Run()
```
karolz-ms, line 72: Consider taking a `CancellationToken` parameter with default value equal to `default` or `CancellationToken.None`

karolz-ms (approved):
👍

src/ApplicationInsights.Kubernetes/K8sInfoBootstrap.cs
xiaomi7732: Done.

Issue #332, opened 2023-01-12
Upgrade KubernetesClient to >= 10.0.16 to remove dependency on System.IO.Abstractions


anhaehne:
With the release of KubernetesClient 10.0.16 the package removed the depency on `System.IO.Abstractions`.
`System.IO.Abstractions` has recently undergone some heavy changes which includes moving the code to a different package under the same namespace. 
More details in my issue in KubernetesClient [Update System.IO.Abstraction to version >= 18.0.0 #1136](https://github.com/kubernetes-client/csharp/issues/1136)

Without removing this dependency, all projects referencing this package can't updated to the latest `System.IO.Abstractions` package without a workaround.

xiaomi7732:
Hi @anhaehne, thanks for filing the issue. Let me take a look.

xiaomi7732:
Hey @anhaehne, I did some investigation, and here are things I am considering bumping `K8s.Client` up to `10.0.16`:

* Minimum target FX .NET 6 (was targeting .NET Standard 2.1)

* K8s cluster version support?
    * https://github.com/kubernetes-client/csharp/issues/1174#issuecomment-1382591832

* K8sClient API break changes
    * List/Read Pods
    * List ReplicaSet
    * List Deployments
    * List Node

xiaomi7732:
By more investigation, another possibility is to make kubernetes client a submodule, reference the sources + K8s.Basic package since K8s.Basic is still targeting `.NET Standard 2.1`, then we don't need to branch out the version of our package. The problem is then, when K8s.Basic bump up to .NET 6, we then will still have to choose a side.

xiaomi7732:
Side note: Another possible approach is the multi-targeting package.

xiaomi7732:
There are 2 technical proposals that are practical, and they each have pros and cons:

|Approaches | Pros | Cons |
|--|--|--|
| Start 6.x, targeting only .NET 6 + | <ul><li>cleaning implementation, easy debugging</li><li>Download number reveals how many users moved to .NET 6 +</li> | User has to understand which version to pick
|Multi-targeting 6.x and netstd 2.1 (.NET Core 3.1) | Users do not need to pick which version to use explicitly | <ul><li>Less motivation to upgrade</li><li>No telemetry to tell how many users moved on.</li><li>More convoluted code to support different targets.</li></ul>

[...]

PR #333, opened 2023-01-24
Support Multi-Targeting

xiaomi7732:
Address #332.

1. For .NET 6 applications, `K8s.Client.10.0.16+` will be depended on;
2. For .NET 3.1 applications, it will stay on `7.x` for backward compatibility;

This is a short-term solution until .NET Core 3.1 is phased out of the system.

Piggyback on:
* an F5 project to allow quick covering .NET Core 3.1 - under folder /dev/F5WebApi.Net31.
* small bug fixes.

### DIFF
--- a/dev/F5WebApi.Net31/F5WebApi.Net31.csproj
+++ b/dev/F5WebApi.Net31/F5WebApi.Net31.csproj
@@ -1,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+  <ItemGroup>
+    
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.21.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights.Kubernetes" Version="3.0.0-private-*" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+</Project>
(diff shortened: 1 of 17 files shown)


dev/F5WebApi.Net31/F5WebApi.Net31.csproj
```diff
@@ -0,0 +1,13 @@
...
+  <ItemGroup>
+    
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.21.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights.Kubernetes" Version="3.0.0-private-*" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
karolz-ms, line 11: Given that .NET 3.1. is out of support, I would not add this example. For the rest of the review, I assume the sample will be removed and I will not review other files that are related to the sample.

karolz-ms (approved):

dev/F5WebApi.Net31/F5WebApi.Net31.csproj
xiaomi7732: I understand. This sample has a different purpose - it is a manual test tool to help cover `.NET Core 3.1` code branch - considering most of the development will be there for `6.x+`.
But anyway, per our discussion, it probably makes more sense to take the other approach. I'll submit another PR to address the issue.

xiaomi7732:
@karolz-ms Thank you for the review. Per our discussion offline, I am going to abandon this PR. Sorry to waste your time reviewing this.

PR #334, opened 2023-01-26
Application Insights for Kubernetes 6.0

xiaomi7732:
Address #332 

* Bump up the major version to drop support for `.NET Core 3.1` for the newer packages.
  * `.NET Core 3.1` is out of support on 12/13/2023. For more info: <https://dot.net>

The existing 3.x package will be kept for a while so that there's time to mitigate existing projects to `6.x`. We will check the download data to determine when to fully remove `Microsoft.ApplicationInsights.Kubernets 3.x`.

### DIFF
--- a/dev/F5WebApi/F5WebApi.csproj
+++ b/dev/F5WebApi/F5WebApi.csproj
@@ -11,7 +11,6 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="6.0.1" />
     <PackageReference Include="Swashbuckle.AspNetCore" Version="6.2.3" />
     <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.*" />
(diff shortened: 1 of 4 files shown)


dev/F5WebApi/F5WebApi.csproj
```diff
@@ -11,6 +11,7 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="6.0.1" />
     <PackageReference Include="Swashbuckle.AspNetCore" Version="6.2.3" />
     <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.*" />
+    <PackageReference Include="KubernetesClient" Version="10.0.16" />
```
xiaomi7732, line 14: This change is not needed. Remove.

xiaomi7732: Removed.

karolz-ms (approved):

PR #337, opened 2023-02-11
Support In-Proc Azure Function in K8s

xiaomi7732:
There is a limitation that there can't be any HostedService registered in the dependency injection container in in-proc Azure Function.

This is an update to:
1. Allow the user to choose to skip registering the BackendService.
2. Expose `IK8sInfoBootstrap` as an alternative to allow the user to invoke it manually.

I would need to provide documentation in the future for the usage.

Tagging @mavsankar since he's asking for the feature.

### DIFF
--- a/src/ApplicationInsights.Kubernetes/Extensions/ApplicationInsightsExtensions.cs
+++ b/src/ApplicationInsights.Kubernetes/Extensions/ApplicationInsightsExtensions.cs
@@ -19,14 +19,16 @@ namespace Microsoft.Extensions.DependencyInjection
         /// </summary>
         /// <param name="services">Collection of service descriptors.</param>
         /// <param name="diagnosticLogLevel">Sets the diagnostics log levels for the enricher.</param>
-        /// <param name="skipRegisterBackendService">Allows skipping register any backend service for constrained environment like InProc Azure Function.</param>
+        /// <param name="disableBackgroundService">If true, the Application Insights enricher library will not use any background (hosted) services,
+        /// and Kubernetes information will not be fetched automatically. Hosted services are not allowed in some environments, e.g. Azure Function. 
+        /// For more information see https://github.com/Azure/azure-functions-host/issues/5447#issuecomment-575368316</param>
         /// <param name="applyOptions">Action to customize the configuration of Application Insights for Kubernetes.</param>
         /// <param name="clusterCheck">Provides a custom implementation to check whether it is inside kubernetes cluster or not.</param>
         /// <returns>The service collection for chaining the next operation.</returns>
         public static IServiceCollection AddApplicationInsightsKubernetesEnricher(
             this IServiceCollection services,
             LogLevel? diagnosticLogLevel = LogLevel.None,
-            bool skipRegisterBackendService = false,
+            bool disableBackgroundService = false,
             Action<AppInsightsForKubernetesOptions>? applyOptions = default,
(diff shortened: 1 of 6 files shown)


karolz-ms:
@xiaomi7732 is the "K8sClient" submodule addition intentional?

src/ApplicationInsights.Kubernetes/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -18,14 +18,16 @@ public static partial class ApplicationInsightsExtensions
         /// Enables Application Insights for Kubernetes on the Default TelemetryConfiguration in the dependency injection container with custom options.
         /// </summary>
         /// <param name="services">Collection of service descriptors.</param>
-        /// <param name="applyOptions">Action to customize the configuration of Application Insights for Kubernetes.</param>
         /// <param name="diagnosticLogLevel">Sets the diagnostics log levels for the enricher.</param>
+        /// <param name="skipRegisterBackendService">Allows skipping register any backend service for constrained environment like InProc Azure Function.</param>
```
karolz-ms, line 22: The fact that the "registration" for backend services is "skipped" is an implementation detail. Consider `disableBackgroundServices` as a name for this parameter.

The description could then say something along the lines of the following (emphasizing user impact):

> If true, the Application Insights enricher library will not use any background (hosted) services,, and Kubernetes information will not be fetched automatically. Hosted services are not allowed in some environments, e.g. Azure Function. For more information see https://relevant/wiki/page.md

karolz-ms (changes requested):

src/ApplicationInsights.Kubernetes/Extensions/ApplicationInsightsExtensions.cs
xiaomi7732: Agree on the naming of the parameter.

I understand the advantages of the auto-detect azure function, here are the considerations:

* It is still not fully automatic at this point - the user will still need to write some other code to bootstrap the detector on their own. And probably a document is still needed.

I say we do it in 2 stages - use this to unblock the scenario. Use another PR in the future to fully automate the scenario - I still can't find a good place to fun background service in Azure Function at this moment.

What do you think?

xiaomi7732:
> @xiaomi7732 is the "K8sClient" submodule addition intentional?

What? Did I append the submodule? Totally a Friday afternoon thingy (it was a total mistake.)

src/ApplicationInsights.Kubernetes/Extensions/ApplicationInsightsExtensions.cs
xiaomi7732: Renamed the parameter name

karolz-ms (approved):

ggirard07:
@xiaomi7732 the link to the repository you mentioned in for isolated model seems broken [in wiki](https://github.com/microsoft/ApplicationInsights-Kubernetes/wiki/%5BAzure-Function-Support%5D-Using-Application-Insights-for-Kubernetes-in-Azure-Functions-hosted-in-K8s-cluster). Is it possible the repo is configured as private on your side? Thanks!

xiaomi7732:
@ggirard07 Thanks for the feedback. Yes, the repo was private. I have made it public:

<https://github.com/xiaomi7732/AzureFuncIsoDocker>

Issue #344, opened 2023-03-23
No container id found by container id providers after AKS upgrade 1.23->1.25


TTmaister:
After upgrade AKS 1.23.x to 1.25.x Application Insight Kubernetes addon can't get container Id if multiple container are used.

RBAC roles and binding has been same before upgrade. 

ContainerIdProviders not 

> [Warning] [2023-03-23T14:50:26.2712637Z] [DockerEngineMountInfoContainerIdProvider] Can't figure out container id.
> [Warning] [2023-03-23T14:50:26.2712637Z] [DockerEngineMountInfoContainerIdProvider] Can't figure out container id.
> [Information] [2023-03-23T14:50:26.2712968Z] No container id found by container id providers.
> [Information] [2023-03-23T14:50:26.2712968Z] No container id found by container id providers.

> # cat /proc/self/mountinfo
> 5504 4910 0:571 / / rw,relatime master:1369 - overlay overlay rw,lowerdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/870/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/727/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/722/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/721/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/720/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/719/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/642/fs,upperdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/871/fs,workdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/871/work
> 5505 5504 0:573 / /proc rw,nosuid,nodev,noexec,relatime - proc proc rw
> 5506 5504 0:574 / /dev rw,nosuid - tmpfs tmpfs rw,size=65536k,mode=755,inode64
> 5507 5506 0:575 / /dev/pts rw,nosuid,noexec,relatime - devpts devpts rw,gid=5,mode=620,ptmxmode=666
> 5508 5506 0:431 / /dev/mqueue rw,nosuid,nodev,noexec,relatime - mqueue mqueue rw
[...]

PR #345, opened 2023-03-28
Container id shall be optional

xiaomi7732:
Trying to address #344 

This is a regression that when there are more than 1 container, and the container id can't be fetched, no telemetry enhancement.

That doesn't agree with the design intention that container id to be optional.

Fix by allowing container id to be absent but other telemetry enhancement still happens.

Here's an example of the logs after the fix:

```shell
[Warning] [2023-03-28T04:18:44.8583934Z] Can't fetch container id. Container id info will be missing. Please file an issue at https://github.com/microsoft/ApplicationInsights-Kubernetes/issues.
[Debug] [2023-03-28T04:18:44.9234011Z] {"ContainerID":null,"ContainerName":null,"PodName":"ai-k8s-f5webapi-deployment-77dc74c9dc-p75vq","PodID":"380deca3-e9c1-405f-b5b0-8e22660a8942","PodLabels":"app:ai-k8s-f5webapi,pod-template-hash:77dc74c9dc","PodNamespace":"ai-k8s-demo","ReplicaSetUid":"87f84c9c-55d2-47d5-82c1-dd6728454b25","ReplicaSetName":"ai-k8s-f5webapi-deployment-77dc74c9dc","DeploymentUid":"c1de6986-b3b1-4f98-b898-008246c0dff5","DeploymentName":"ai-k8s-f5webapi-deployment","NodeName":"docker-desktop","NodeUid":"cd59d76e-eaf2-4408-9083-aa2d76548055"}
```

### DIFF
--- a/src/ApplicationInsights.Kubernetes/Containers/ContainerStatusManager.cs
+++ b/src/ApplicationInsights.Kubernetes/Containers/ContainerStatusManager.cs
@@ -1,5 +1,7 @@
 using System;
+using System.Collections.Generic;
 using System.Net;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


karolz-ms:
@xiaomi7732 and I had a conversation about this. We agreed it will be better to stop requiring container readiness before container info is fetched from the API server. A new PR (or substantial revision to this PR) is in the works.

xiaomi7732:
See #348

Issue #355, opened 2024-01-02
Can't register Enrichers when using KeyedServices


tomaszzmuda:
**Describe the bug**
Can't register Enrichers when using KeyedServices when using .NET 8

**Package Versions**
Application Insights Kubernetes Version: 6.1.1
Application Insights SDK Version: 2.22.0

**To Reproduce**
1. Create .NET 8 application with dependency injection and keyed services
2. Try to invoke **AddApplicationInsightsKubernetesEnricher**
3. Get an exception

**Expected behavior**
It is possible to add Kubernetes enrichers when you used keyed services.

**Screenshots**
Stacktrace:
![image](https://github.com/microsoft/ApplicationInsights-Kubernetes/assets/13612984/47255677-88ea-44aa-b8e1-e1c96c1621ab)

**Additional context**

https://github.com/AzureAD/microsoft-identity-web/issues/2604

In **AddApplicationInsightsKubernetesEnricher** method there is a check if initalizer already exists: **KubernetesTelemetryInitializerExists**. Implementation of that method checks if any ImplementationType from service collection is of type KubernetesTelemetryInitializer but getting ImplementationType property by design throws exception if it's keyed service.

```c#
/// <summary>
/// Gets the <see cref="Type"/> that implements the service.
/// </summary>
[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)]
public Type? ImplementationType
{
    get
    {
        if (IsKeyedService)
        {
            ThrowKeyedDescriptor();
        }
        return _implementationType;
    }
}
``` 

```c#
//
// Summary:
//     Checks if the KubernetesTelemetryInitializer exists in the service collection.
//
//
// Parameters:
//   serviceCollection:
//     The service collection.
private static bool KubernetesTelemetryInitializerExists(IServiceCollection serviceCollection)
{
    return serviceCollection.Any((ServiceDescriptor t) => t.ImplementationType == typeof(KubernetesTelemetryInitializer));
}
```

xiaomi7732:
Beta package is released, please give it a try.
https://www.nuget.org/packages/Microsoft.ApplicationInsights.Kubernetes/6.1.2-beta1

PR #356, opened 2024-01-05
Bugfix/355 - Migration to .NET 8

tomaszzmuda:
Solves that bug: https://github.com/microsoft/ApplicationInsights-Kubernetes/issues/355
It could have been done without upgrade but it would be just a temporary fix.

### DIFF
--- a/src/ApplicationInsights.Kubernetes/ApplicationInsights.Kubernetes.csproj
+++ b/src/ApplicationInsights.Kubernetes/ApplicationInsights.Kubernetes.csproj
@@ -19,12 +19,12 @@
 
   <ItemGroup>
     <PackageReference Include="KubernetesClient" Version="[10.0.16, 11.0.0)" />
-    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.21.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.22.0" />
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="8.0.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="8.0.0" />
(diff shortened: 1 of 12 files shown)


src/ApplicationInsights.Kubernetes/ApplicationInsights.Kubernetes.csproj
```diff
@@ -3,7 +3,7 @@
     <Features>IOperation</Features>
   </PropertyGroup>
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
```
xiaomi7732, line 6: Bumping up the target to net8 would require all the apps upgrade to .NET 8, which is a major backward compatibility concerning.

It is a bit too early considering the rest of lifespan left for .NET 6:
image

It is mentioned that `It could have been done without upgrade`, I am wondering what alternatives are there? Do you mind try something without bumping up to .NET 8 (yet).

xiaomi7732 (changes requested):

src/ApplicationInsights.Kubernetes/ApplicationInsights.Kubernetes.csproj
xiaomi7732: How about just bumping up the package version of `Microsoft.Extensions.DependencyInjection` to 8.0.0? It is easier, and it supports `.net 6` and above.

Notice, technically, we could just reference `Microsoft.Extensions.DependencyInjection.Abstractions`, but that would lead to runtime exception like the one reported for .NET 6 applications when the referenced `Microsoft.Extensions.DependencyInjection` is older. To fix it, the user would have to reference of the higher version of it anyway.

xiaomi7732:
Considering to have PR #360 to replace this one.

xiaomi7732:
The other PR is in.

Issue #357, opened 2024-01-09
Upgrade of `KubernetesClient` dependency


tomachristian:
**Is your feature request related to a problem? Please describe.**
We are using `Microsoft.Orleans.Hosting.Kubernetes 8.0.0` which uses `KubernetesClient (>= 12.1.1)` whilst latest `Microsoft.ApplicationInsights.Kubernetes 6.1.1` is using `KubernetesClient (>= 10.0.16 && < 11.0.0)`. For this reason we cannot use them together. Any plans to upgrade the `KubernetesClient` dependency?

**Describe the solution you'd like**
That the `KubernetesClient` be upgraded to the latest version.

**Describe alternatives you've considered**
We couldn't find any besides dropping this library completely.

tomachristian:
@xiaomi7732 any timeline on when these changes will be released?

xiaomi7732:
Hey @tomachristian I will work on a beta package for testing in a couple of days. Will that work for you?

xiaomi7732:
Please check this one: https://www.nuget.org/packages/Microsoft.ApplicationInsights.Kubernetes/6.1.2-beta1

nZeus:
Thank you @xiaomi7732 ! 👍

dsempi:
@xiaomi7732: Can we reopen the issue? There were breaking changes between KubernetesClient 10 and 11. Specifically, `KubernetesClient.Models` and `KubernetesClient.Basic` are no longer separate and are now part of the `KubernetesClient` assembly itself. Could we please have a version that references the newer version of the `KubernetesClient` assembly? Thank you!

xiaomi7732:
@dsempi, thanks for sharing the details. That's kind of issue that would happen upon losing the dependencies. I could bump up the client to use K8sClient 11 or even 12, but there's still potential problems for 13 and 14 in the future.
It looks like we will have to have a strategy to depend on various versions.

One way is to sim-ship multiple versions, the problem is that it will greatly increase the maintains cost.

Maybe a way out is to ship a major version that maps to the k8s client major versions, and then, keep up with the latest?

For example, the current 2.x will keep using 10.x; and then, release a 11.0 to use 11; release a 12.0 to use 12; release a 13.0 to use 13. Moving forward, we will update the dependencies gradually only for 13. And hopefully, when 14 releases, we will be ready to move on to use 14 already? Will that work?

What do we think will be a good version to dependent on at this point?

[...]

PR #358, opened 2024-01-09
Upgrade KubernetesClient dependency

tomachristian:
Fixes https://github.com/microsoft/ApplicationInsights-Kubernetes/issues/357.

### DIFF
--- a/src/ApplicationInsights.Kubernetes/ApplicationInsights.Kubernetes.csproj
+++ b/src/ApplicationInsights.Kubernetes/ApplicationInsights.Kubernetes.csproj
@@ -18,7 +18,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="KubernetesClient" Version="[10.0.16, 11.0.0)" />
+    <PackageReference Include="KubernetesClient" Version="10.0.16" />
     <PackageReference Include="Microsoft.ApplicationInsights" Version="2.21.0" />
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="6.0.1" />
     <PackageReference Include="Microsoft.Extensions.Hosting" Version="6.0.1" />


didranoqx:
Hi, this commit is important for me. Have you some ideas when it could be approved ?

tomachristian:
Hey @xiaomi7732, sorry to mention you directly, but do you think you could help us with this?

src/ApplicationInsights.Kubernetes/ApplicationInsights.Kubernetes.csproj
```diff
@@ -18,7 +18,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="KubernetesClient" Version="[10.0.16, 11.0.0)" />
+    <PackageReference Include="KubernetesClient" Version="12.1.1" />
```
xiaomi7732, line 21: This bumps up the lowest version supported.
If I understand correct, the version needs to be uncapped to unblock the bumping up of the dependencies. How about make it `10.0.16`? That would be equivalent to `[10.0.16, )`.

Considerations:
1. Backward compatibility for existing code relying on the lower version;
2. Potentially breaking changes introduced in the K8sClient (major version upgrades);

If that would work for you, could you please update the PR?

xiaomi7732 (changes requested):

src/ApplicationInsights.Kubernetes/ApplicationInsights.Kubernetes.csproj
tomachristian: I upgraded the reference as per your request. Thank you!

xiaomi7732 (approved):
LGTM. Thanks for the contribution!

src/ApplicationInsights.Kubernetes/ApplicationInsights.Kubernetes.csproj
xiaomi7732: Cool!

Issue #369, opened 2024-07-30
Stop overwrite sdk version


xiaomi7732:
It doesn't make sense to have this enhancer to carry a separate SDK version. That actually created quite a confusion on the support side.

For the backward compatibility, we should also expose a switch to allow keep the sdk version.

For example:

```
KeepUsingCustomSDKVersion = `true` | `false` and default to `false`.
```

Break down the tasks:

[ ] Stop overwriting sdk version;
[ ] Add an option to allow the end user choose to keep writing the sdk version;
[ ] Update document about the configuration.

xiaomi7732:
Package released: https://www.nuget.org/packages/Microsoft.ApplicationInsights.Kubernetes

PR #381, opened 2025-03-08
Dev/saars/fix stop overwriting sdk version

xiaomi7732:
Addresses #369

### DIFF
--- a/src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
+++ b/src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
@@ -5,6 +5,8 @@ using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.ApplicationInsights.Extensibility.Implementation;
 using Microsoft.ApplicationInsights.Kubernetes.Debugging;
 using Microsoft.ApplicationInsights.Kubernetes.Utilities;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
 using static Microsoft.ApplicationInsights.Kubernetes.TelemetryProperty;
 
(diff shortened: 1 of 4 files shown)


src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
```diff
@@ -16,15 +18,19 @@ internal class KubernetesTelemetryInitializer : ITelemetryInitializer
...
+
     internal IK8sEnvironmentHolder K8SEnvironmentHolder { get; }
     internal ITelemetryKeyCache TelemetryKeyCache { get; }
 
     public KubernetesTelemetryInitializer(
         IK8sEnvironmentHolder k8sEnvironmentHolder,
         SDKVersionUtils sdkVersionUtils,
-        ITelemetryKeyCache telemetryKeyCache)
+        ITelemetryKeyCache telemetryKeyCache,
+        IOptions<AppInsightsForKubernetesOptions> options)
```
karolz-ms, line 30: It is an internal class so it does not matter that much, but if an uninitialized instance of options is valid, you could accept a null value and internally treat it the same as uninitialized instance.

karolz-ms (approved):

src/ApplicationInsights.Kubernetes/TelemetryInitializers/KubernetesTelemetryInitializer.cs
xiaomi7732: @karolz-ms, I don't think I understand what you are saying. Do you mind elaborate a bit? Like give me an example? Are you saying I could have accepted options to be null, instead of throwing?

karolz-ms: Let's have a coffee sometime and I will explain  😆

PR #384, opened 2025-08-08
Ability to exclude node information

okogut:
Some companies do not allow all pods to call the nodes api in the cluster, 

Currently there is no way to skip the call to the nodes endpoint, which if you do not have the permission will potentially make noise in the logs showing a lot of failed (403 Forbidden) HTTP requests (see below). 
`K8sClientService.GetNodesAsync` ignores the forbidden response, by returning an empty `IEnumerable<V1Node>`, but the request is still being performed.

This change adds an option to `ExcludeNodeInformation`, which you can set to `true`, if you want to skip the call to nodes endpoint entirely - typically because you don't have permission (and likely won't get it) to call the endpoint.

image


okogut:
@xiaomi7732 I updated what you mentioned.

xiaomi7732 (approved):
LGTM

okogut:
A test failed due to missing setup of `IOptions<AppInsightsForKubernetesOptions>.Value` - fixed now.

xiaomi7732 (approved):
LGTM

xiaomi7732:
Will be released in 8.0.0-beta2.

PR #395, opened 2026-09-21
Remove indirect eval from example jQuery files

cweining:
The built-in eval function and the Function constructor allow executing arbitrary strings as JavaScript code. This is a dangerous feature, since this code has the same access privileges as any other code, so great care has to be taken to ensure that malicious code is not accidentally executed this way. Using this feature also hampers static checking and program comprehension. In many cases, better alternatives are available and should be used instead.


xiaomi7732 (approved):
LGTM.

Voyagerroc-Lab (approved):
Review summary

The indirect-eval branch is removed in all three checked-in example copies, while the existing script-injection path is retained for execution. The builds for both debug and release passed, and the change is limited to vendored example jQuery files.

I found no blocking issue in this focused security cleanup. Approving.

